Issue #6, opened 2015-06-17
`preexec` doesn't work for functions


evverx:
```
$ source .bash-preexec.sh
$ preexec() { echo "just typed $1"; }
$ date
just typed date
Wed Jun 17 16:20:15 UTC 2015
$ function foo() { echo foo; }
$ foo
just typed foo
foo
```

rcaloras:
Hey thanks for the feedback! This is a known limitation along with subshell parsing e.g. doing something like 
```bash
(ls)
```

I have a branch in progress for this fix, but it's not too easy. https://github.com/rcaloras/bash-preexec/tree/SubShellParsing.

It actually takes a decent amount of time for bash to handle this case and I believe it creates some lag. Would love your contribution if you can help figure it out. I'll give it another look tonight as well.

evverx:
@rcaloras , thanks. I'll read https://github.com/rcaloras/bash-preexec/tree/SubShellParsing this evening:)

evverx:
Reminder for me:
```
$ source .bash-preexec.sh
$ preexec() { echo "just typed $1"; }
$ date
just typed date
Wed Jun 17 17:07:28 UTC 2015
$ <SPACE>who
just typed date
vagrant  pts/1        2015-06-17 17:06 (10.0.2.2)
```

rcaloras:
That should address the space issue :) Not sure about functions yet though.

evverx:
@rcaloras , thanks:)
But `man bash` says:
```
A value of ‘ignoreboth’ is shorthand for ‘ignorespace’ and ‘ignoredups’
```
Overwriting `HISTCONTROL` is privacy issue:)

rcaloras:
Yeah it should also replace `ignoreboth` with just `ignoredups`.

Why is overwriting histcontrol a privacy issue? This plugin overwrites both bash's  `DEBUG` trap and `PROMPT_COMMAND` IMO that's more invasive :P

evverx:
that's the joke:)

Reminder for me:
Some useful variables and options:
```
     set -o history enable command history

     HISTSIZE
              The number of commands to remember in the command  history.
              If  the value is 0, commands are not saved in
              the history list.  Numeric values less than zero result in every
              command  being  saved  on  the history list (there is no limit).
              The shell sets the  default  value  to  500  after  reading  any
              startup files.

     HISTIGNORE
              A colon-separated list of patterns used to decide which  command
              lines  should  be  saved  on  the history list.
```

[...]

PR #8, opened 2015-06-23
Add HISTTIMEFORMAT support

evverx:
See
https://github.com/rcaloras/bash-preexec/issues/6#issuecomment-114007154


rcaloras:
Amazing :+1: 

Thanks so much!

PR #9, opened 2015-06-28
Invoke preexec function for subshells and funtions #6

rcaloras:
- Watches for history changes to properly invoke preexec
  function. Handles duplicates and shell startup edge cases.

- Still a bit of a hack since it's not truly prexection (i.e. will come after
  the function or subshell is executed) but the only way I
  found in bash to get prexec to invoke for both functions and subshells.

### DIFF
--- a/bash-preexec.sh
+++ b/bash-preexec.sh
@@ -123,47 +123,54 @@ __bp_in_prompt_command() {
     return 1
 }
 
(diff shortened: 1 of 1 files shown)


rcaloras:
Closed in favor of https://github.com/rcaloras/bash-preexec/pull/22

PR #12, opened 2015-10-11
fix localized loop var typo

calid:
very minor, the loop var wasn't actually being unset because of a name mismatch


rcaloras:
Nice catch, thanks for the PR! Much appreciated sir :D 

Check out my other project [Bashhub](https://github.com/rcaloras/bashhub-client), that bash-preexec supports. Would love to hear your feedback.

rcaloras:
:+1:  :hamburger:  :fries:

PR #32, opened 2016-12-03
Don't clobber $_

HaleTom:
`$_` should give the last argument of the last command. 

Currently it gives `__bp_preexec_invoke_exec`:
```
$ exec bash
$ . ./bash-preexec.sh
$ preexec() { echo "just typed $1"; }
$ echo $_
just typed echo $_
__bp_preexec_invoke_exec
$
```

Fix this to produce the expected behaviour:
```
exec bash
$ . ./bash-preexec.sh
$ preexec() { echo "just typed $1"; }
$ echo $_
just typed echo $_
./bash-preexec.sh
```

Note that `bash-bats` currently suffers from the *same bug* (it's a **feature** of bash), so the newly added test fails. I have manually tested it and it passes.

PR #35, opened 2017-04-13
Preserve return value of last function in preexec()

brandonweeks:
This change causes `__bp_preexec_invoke_exec()` to return with the value from the last function in `preexec()`.  As of Bash 3.0 if `extdebug` is set a non-zero return value will cause the next command to be skipped.

http://git.savannah.gnu.org/cgit/bash.git/tree/NEWS#n1001

### DIFF
--- a/bash-preexec.sh
+++ b/bash-preexec.sh
@@ -191,6 +191,7 @@ __bp_preexec_invoke_exec() {
 
     # For every function defined in our function array. Invoke it.
     local preexec_function
+    local preexec_ret_value
     for preexec_function in "${preexec_functions[@]}"; do
 
         # Only execute each function if it actually exists.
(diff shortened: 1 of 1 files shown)


rcaloras:
@brandonweeks thanks for the PR! Quick check, does this interfere with the `$__bp_last_argument_prev_command` at all? e.g. you should still be able to use `$_` which was setup in https://github.com/rcaloras/bash-preexec/commit/0c2f061e691241972abf42ccaff954625bd5f01d

brandonweeks:
Attempting to test this, even without this PR `$_` doesn't seem to work for me in `precmd()`.

brandonweeks:
Apparently `shopt -s extdebug` breaks it, works otherwise.

rcaloras:
Yeah extdebug causes some instability over all. It's the key to subshell support though. It's disabled by default in bash-preexec https://github.com/rcaloras/bash-preexec#subshells
https://github.com/rcaloras/bash-preexec/blob/master/bash-preexec.sh#L255

rcaloras:
So long as it doesn't break `$_` when extdebug is not enabled. An example is in https://github.com/rcaloras/bash-preexec/pull/32. Otherwise this LGTM.

brandonweeks:
Refactored to use the `__bp_set_ret_value` function instead of setting the `$_` and `$?` separately.

```
bash-4.4$ echo "hello"
hello
precmd $_: hello
precmd $?: 0
bash-4.4$ 
```
`$_` still working as expected unless `shopt -s extdebug` is ran.

rcaloras:
LGTM :+1: :hamburger: :fries:

rcaloras:
@brandonweeks thanks for your contributions!

Issue #36, opened 2017-04-27
injecting preexec function name into terminal when using editor


CeleritasCelery:
When editing a Bash command you can use `Ctrl-x Ctrl-e` to open the command in `$EDITOR`, which I have set to Vim. Once I am doing editing the command, i execute `:wq` to quit Vim and when I look at my terminal I see `__bp_preexec_invoke_exec` being injected into the output.

So for example if I create the command `echo foo` in vim then ran it, my output would look like 
```
> echo foo
__bp_preexec_invoke_exec
foo
```

Any idea why this might be happening and is there a fix?

I am using bash 4.4

D630:
Maybe you can define a preexec function with something like `echo "$BASH_COMMAND"` to see what is beeing executed by "edit-and-execute-command".

I get the follwowing:
```sh
fc -e "${VISUAL:-${EDITOR:-$(command -v editor || echo emacs)}}"
/usr/bin/vim /run/user/1000/tmp/bash-fc.IUxsw9
```

CeleritasCelery:
when I use set -v I get and try the command `echo foo`
```
fc -e "${VISUAL:-${EDITOR:-emacs}}"
__bp_preexec_invoke_exec
vim /tmp/bash-fc-9053782316985282392
__bp_preexec_invoke_exec
echo foo
__bp_preexec_invoke_exec
foo
```

Does it not inject the function into the output for you?

rcaloras:
@CeleritasCelery Yes. I'm getting the same output as in your top example. Wondering what's causing it in bash-preexec.

dschep:
I've noticed this behavior for a while but I just ignore it :stuck_out_tongue:

rcaloras:
I'm not positive there's anything that can be done about this since bash-preexec relies on bash's `DEBUG` trap.

Even setting a no-op here will be echoed out. Using your example from above @CeleritasCelery  e.g.
```bash
$ trap -- ':' DEBUG
Ctrl-x Ctrl-e
echo foo
:
foo
```

rcaloras:
@CeleritasCelery unfortunately this is a product of using the `DEBUG` trap which is pretty much a must.
@D630 any idea why edit and execute pulls in the contents of the trap?

D630:
@rcaloras It's not because of Readline's "edit-and-execute-command", it's due to fc, that
sets the verbose flag before executing the file. See
[...]

PR #38, opened 2017-05-08
Initalize preexec_ret_value to prevent error

brandonweeks:
Fixing a bug I introduced in #36. If `preexec_functions` was empty, the following error would occur.

```sh
bash-4.4$ source bash-preexec.sh 
bash-4.4$ true
bash: return: bash-preexec.sh: numeric argument required
bash-4.4$ 
```

I wasn't able to find a way to write a test for this behavior in Bats, if you have any idea how to observe the error within Bats let me know.

### DIFF
--- a/bash-preexec.sh
+++ b/bash-preexec.sh
@@ -191,7 +191,7 @@ __bp_preexec_invoke_exec() {
 
     # For every function defined in our function array. Invoke it.
     local preexec_function
-    local preexec_ret_value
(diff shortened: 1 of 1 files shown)


rcaloras:
@brandonweeks Thanks for fixing. :+1: :hamburger: :fries:

CeleritasCelery:
I think you meant to reference #35 (which is your PR) not #36 (which is an unrelated issue)

PR #41, opened 2017-05-09
Remove trailing semicolon to allow appending to PROMPT_COMMANDS

brandonweeks:
Before:
```sh
$ source bash-preexec.sh
$ PROMPT_COMMAND="$PROMPT_COMMAND; true"
bash: PROMPT_COMMAND: line 3: syntax error near unexpected token `;;'
bash: PROMPT_COMMAND: line 3: `__bp_precmd_invoke_cmd;  __bp_interactive_mode;; true'
$ 
```

After:
```sh
$ source bash-preexec.sh
$ PROMPT_COMMAND="$PROMPT_COMMAND; true"
$ declare -p PROMPT_COMMAND
declare -- PROMPT_COMMAND="__bp_precmd_invoke_cmd;  __bp_interactive_mode; true"
$ 
```

### DIFF
--- a/bash-preexec.sh
+++ b/bash-preexec.sh
@@ -267,7 +267,7 @@ __bp_install() {
 
     # Install our hooks in PROMPT_COMMAND to allow our trap to know when we've
     # actually entered something.
-    PROMPT_COMMAND="__bp_precmd_invoke_cmd; ${existing_prompt_command} __bp_interactive_mode;"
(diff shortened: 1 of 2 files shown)


rcaloras:
This one is more of a design decision right? I feel like you could argue that we should still end `PROMPT_COMMAND` with a semicolon and then just append without it.

rcaloras:
:+1: Putting it through regardless. Thanks for all the PRs! 

@brandonweeks :star: :1st_place_medal: 

alt-text

PR #46, opened 2017-06-25
Change to preexec function error code

jombooth:
Some friends and I are using bash-preexec in an application wherein our users sometimes load multiple projects that depend on bash-preexec in their shells. One of these uses extdebug, along with the useful feature that the command the user tries to run is skipped if the last function in preexec_functions returns an error - but we run into problems where the last function in preexec ends up being, eg, one that always returns 0, because the user also sources another script which tacks an extra function onto the end of preexec_functions.

This script changes the behavior of bash-preexec so that an error from any function in the array causes the overall debug trap call to return 1, so that extdebug can be used reliably in conjunction with multiple projects that use bash-preexec.

### DIFF
--- a/bash-preexec.sh
+++ b/bash-preexec.sh
@@ -91,7 +91,7 @@ __bp_precmd_invoke_cmd() {
     # Save the returned value from our last command
     __bp_last_ret_value="$?"
 
-    # For every function defined in our function array. Invoke it.
+    # Invoke every function defined in our function array.
     local precmd_function
     for precmd_function in "${precmd_functions[@]}"; do
 
@@ -189,8 +189,9 @@ __bp_preexec_invoke_exec() {
     # the command is in fact interactive and we should invoke the user's
     # preexec functions.
 
-    # For every function defined in our function array. Invoke it.
+    # Invoke every function defined in our function array.
     local preexec_function
+    local preexec_function_ret_value
     local preexec_ret_value=0
     for preexec_function in "${preexec_functions[@]}"; do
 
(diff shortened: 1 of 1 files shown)


bash-preexec.sh
```diff
@@ -199,14 +199,16 @@ __bp_preexec_invoke_exec() {
         if type -t "$preexec_function" 1>/dev/null; then
             __bp_set_ret_value $__bp_last_ret_value
             $preexec_function "$this_command"
-            preexec_ret_value="$?"
+            if (( $? != 0 )); then
```
rcaloras: Small style comments. Also why don't we set the value to `$?`?
```
if [[ "$?" != 0 ]]; then
  preexec_ret_value="$?"
fi

File: bash-preexec.sh [hunk 2]
```diff
@@ -199,14 +199,16 @@ __bp_preexec_invoke_exec() {
...
+                preexec_ret_value=1
+            fi
         fi
     done
 
-    # Restore the last argument of the last executed command
-    # Also preserves the return value of the last function executed in preexec
-    # If `extdebug` is enabled a non-zero return value from the last function
-    # in prexec causes the command not to execute
+    # Restore the last argument of the last executed command, and set the return
```
rcaloras: Thanks for updating comments 👍

rcaloras:
@brandonweeks Any thoughts on this change? You were the original contributor of preserving the return status of the last function :)

bash-preexec.sh
brandonweeks: LGTM. This preserves the "blocking" behavior of preexec returning non-zero even if another preexec function returns 0, which is ideal.

rcaloras:
@jombooth I'll stamp and merge once comments are addressed. Thanks for submitting!

bash-preexec.sh
jombooth: Changed to $?.

bash-preexec.sh
jombooth: np! :)

jombooth:
SGTM - of course!

bash-preexec.sh
```diff
@@ -199,14 +200,18 @@ __bp_preexec_invoke_exec() {
         if type -t "$preexec_function" 1>/dev/null; then
             __bp_set_ret_value $__bp_last_ret_value
             $preexec_function "$this_command"
-            preexec_ret_value="$?"
+            preexec_function_ret_value="$?"
+            if (( $preexec_function_ret_value != 0 )); then
```
rcaloras: Can you also change this to be double square brackets? That's the style for this file. 
```
if [[ $preexec_function_ret_value != 0 ]]; then
```

jombooth: The round brackets are actually a different construct (arithmetic expansions - http://tldp.org/LDP/abs/html/arithexp.html#ARITHEXPREF) that seem pretty reasonable here, since the variable used is guaranteed to be an integer. Doing it the other way would be a string comparison, which works, but I think this approach is preferable. Happy to be overruled on this, as well.

jombooth: Also - the reason I didn't quote the variable is that they're not needed inside arithmetic expansions as word splitting, etc, don't occur. The $ could be dropped as well, but I find code that does that a little confusing to read since I'm used to $-dereferencing of variables.

jombooth: Hi - decided to go ahead and make the change.

jombooth:
I made the change requested earlier, from (( to [[. Please let me know if anything else is needed!

bash-preexec.sh
rcaloras: @jombooth I think your point is valid! I would have merged :) This way is more consistent I guess.

rcaloras:
@jombooth Sorry for the delayed response. Thanks for addressing changes and your contributions! Much obliged!

rcaloras:
:+1: :hamburger: :fries:

PR #47, opened 2017-06-26
Refactor storing PROMPT_COMMAND to use a dynamic function

brandonweeks:
First stab at using a function to store PROMPT_COMMAND instead of inside PROMPT_COMMAND itself. If we agree on a solid implementation, I'd like to use the same strategy to store the DEBUG trap.

### DIFF
--- a/bash-preexec.sh
+++ b/bash-preexec.sh
@@ -211,35 +211,7 @@ __bp_preexec_invoke_exec() {
     __bp_set_ret_value "$preexec_ret_value" "$__bp_last_argument_prev_command"
 }
 
-# Returns PROMPT_COMMAND with a semicolon appended
-# if it doesn't already have one.
-__bp_prompt_command_with_semi_colon() {
-
-    # Trim our existing PROMPT_COMMAND
-    local trimmed
-    trimmed=$(__bp_trim_whitespace "$PROMPT_COMMAND")
-
-    # Take our existing prompt command and append a semicolon to it
-    # if it doesn't already have one.
-    local existing_prompt_command
-    if [[ -n "$trimmed" ]]; then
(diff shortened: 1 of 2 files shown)


bash-preexec.sh
```diff
@@ -293,12 +260,13 @@ __bp_install_after_session_init() {
         return 1;
     fi
 
-    local existing_prompt_command
-    existing_prompt_command=$(__bp_prompt_command_with_semi_colon)
+    if [[ -n "$PROMPT_COMMAND" ]]; then
+        eval "original_prompt_command() { $PROMPT_COMMAND; }"
```
rcaloras: Small edge case here. I have a semi colon previously at the end of my prompt command which errors in this version with:

```bash
bash: syntax error near unexpected token `;;'
```
Any ideas to address?

bash-preexec.sh
```diff
@@ -261,18 +232,14 @@ __bp_install() {
...
     # Install our hooks in PROMPT_COMMAND to allow our trap to know when we've
     # actually entered something.
-    PROMPT_COMMAND="__bp_precmd_invoke_cmd; ${existing_prompt_command} __bp_interactive_mode"
+    PROMPT_COMMAND="__bp_precmd_invoke_cmd; __bp_interactive_mode"
     eval "$__bp_trap_install_string"
 
     # Add two functions to our arrays for convenience
     # of definition.
-    precmd_functions+=(precmd)
+    precmd_functions+=(precmd, original_prompt_command)
```
rcaloras: Can you update the comment as well to reflect adding the original prompt command?

rcaloras:
@brandonweeks closing this as it was addressed in https://github.com/rcaloras/bash-preexec/pull/50. Thanks again for the original PR! 👍

PR #49, opened 2017-07-27
Fix two tests in bash-preexec.bats.

dimo414:
* There's no point checking $output if run isn't being called.
* Need to eval the value of $PROMPT_COMMAND separately from updating it.
  See https://github.com/rcaloras/bash-preexec/commit/64c6f53b52#commitcomment-23327269


rcaloras:
LGTM :+1: :hamburger: :fries: 

@dimo414 Thanks for the contributions!

Issue #39, opened 2017-05-09
Preserve existing DEBUG traps 


brandonweeks:
Currently bash-preexec clobbers existing DEBUG traps when it is sourced. Would you be interested in a PR that moves the contents of `trap -p DEBUG` into a function and appends it to `preexec_functions`?

dimo414:
Similar support for `PROMPT_COMMAND` (appended to `precmd_functions`) would also be nice.

rcaloras:
the original `PROMPT_COMMAND` should be preserved as part of `PROMPT_COMMAND`still. 
https://github.com/rcaloras/bash-preexec/blob/master/bash-preexec.sh#L270

We could also move it to the precmd_functions array as well.

dimo414:
Ah, nice. Moving it to `precmd_functions` (if that's what's done for the `DEBUG` trap) would make sense assuming it wouldn't break anyone. Otherwise leaving it as-is is probably fine.

dimo414:
> Currently bash-preexec clobbers existing DEBUG traps when it is sourced.

I'm actually not sure this is true; I found that [sourced scripts can't override an existing DEBUG trap](https://stackoverflow.com/q/43989793/113632). Worse, it seems from within the sourced script there's no way to tell whether a DEBUG trap was already set or not. I'm not sure if this FR is possible.

rcaloras:
@dimo414 I noticed this a little while back. The debug trap isn't actually set when sourcing the script. bash-preexec actually does a two part install where it hooks its install function first into `PROMPT_COMMAND` and then runs setup through the invocation of that function. e.g. https://github.com/rcaloras/bash-preexec/blob/master/bash-preexec.sh#L306. and https://github.com/rcaloras/bash-preexec/blob/master/bash-preexec.sh#L241

Original commit for it: https://github.com/rcaloras/bash-preexec/commit/4744d7e6fd105eff8bdcb2ede07db603c5809c88

dimo414:
Well that's clever :) It would not have crossed my mind to try that. In that case it seems like it should be possible to add the trap to `preexec_functions` with something like:

    preexec_functions+=("$(trap -p DEBUG | sed -e "s/trap -- '//" -e "s/' DEBUG//")")

There's probably a better way to extract the DEBUG command, but that worked in a pinch. I take it your open to a pull request adding something along these lines?

rcaloras:
Of course! Please feel free  to submit a PR :+1:

[...]

PR #50, opened 2017-07-31
Ensure a pre-existing DEBUG trap is preserved as a preexec function.

dimo414:
* Capture any pre-existing PROMPT_COMMAND hook in the same way, adding it as a precmd function.
* No longer need to do string manipulations of the PROMPT_COMMAND's contents.
* Added a unit test of the new trapping semantics.
* Tidied up some other tests to test the public API and avoid tweaking internal state
  (e.g. the old $_ test would fail to catch a bug in this change).

Resolves issue #39.

### DIFF
--- a/bash-preexec.sh
+++ b/bash-preexec.sh
@@ -26,13 +26,13 @@
 #
 #       preexec_functions+=(my_preexec_function)
 #
-#  3. If you have anything that's using the Debug Trap, change it to use
-#     preexec. (Optional) change anything using PROMPT_COMMAND to now use
-#     precmd instead.
+#  3. Consider changing anything using the DEBUG trap or PROMPT_COMMAND
+#     to use preexec and precmd instead. Preexisting usages will be
+#     preserved, but doing so manually may be less surprising.
 #
-#  Note: This module requires two bash features which you must not otherwise be
-#  using: the "DEBUG" trap, and the "PROMPT_COMMAND" variable. prexec_and_precmd_install
-#  will override these and if you override one or the other this will most likely break.
+#  Note: This module requires two Bash features which you must not otherwise be
+#  using: the "DEBUG" trap, and the "PROMPT_COMMAND" variable. If you override
+#  either of these after bash-preexec has been installed it will most likely break.
 
 # Avoid duplicate inclusion
 if [[ "$__bp_imported" == "defined" ]]; then
@@ -44,10 +44,6 @@ __bp_imported="defined"
 # functions, should they want it.
 __bp_last_ret_value="$?"
 __bp_last_argument_prev_command="$_"
-
-# Command to set our preexec trap. It's invoked once via
-# PROMPT_COMMAND and then removed.
-__bp_trap_install_string="trap '__bp_preexec_invoke_exec \"\$_\"' DEBUG;"
 
 # Remove ignorespace and or replace ignoreboth from HISTCONTROL
 # so we can accurately invoke preexec with a command from our
@@ -137,8 +133,6 @@ __bp_in_prompt_command() {
 # environment to attempt to detect if the current command is being invoked
 # interactively, and invoke 'preexec' if so.
 __bp_preexec_invoke_exec() {
-
-
     # Save the contents of $_ so that it can be restored later on.
     # https://stackoverflow.com/questions/40944532/bash-preserve-in-a-debug-trap#40944702
     __bp_last_argument_prev_command="$1"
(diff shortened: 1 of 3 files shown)


dimo414:
Travis reports "The command "sudo apt-get install -qq bats" failed and exited with 100 during ."  - seems like a CI bug? All tests passed locally, FWIW.

dimo414:
@rcaloras just checking if you've had a chance to look at this PR?

bash-preexec.sh
```diff
@@ -47,7 +47,17 @@ __bp_last_argument_prev_command="$_"
...
 # Command to set our preexec trap. It's invoked once via
 # PROMPT_COMMAND and then removed.
-__bp_trap_install_string="trap '__bp_preexec_invoke_exec \"\$_\"' DEBUG;"
+__bp_trap_install() {
+  local prior_trap=$(trap -p DEBUG | sed "s/[^']*'\(.*\)'[^']*/\1/")
+  if [[ -n "$prior_trap" ]]; then
+    eval '__bp_original_debug_trap() {
+      '"$prior_trap"'
+    }'
+    preexec_functions+=("$__bp_original_debug_trap")
```
rcaloras: Looks like the way you're adding to preexec functions is different than precmd. Shouldn't this be like:

```bash
preexec_functions+=(__bp_original_debug_trap)
```

bash-preexec.sh
```diff
@@ -47,7 +47,17 @@ __bp_last_argument_prev_command="$_"
 
 # Command to set our preexec trap. It's invoked once via
 # PROMPT_COMMAND and then removed.
-__bp_trap_install_string="trap '__bp_preexec_invoke_exec \"\$_\"' DEBUG;"
+__bp_trap_install() {
+  local prior_trap=$(trap -p DEBUG | sed "s/[^']*'\(.*\)'[^']*/\1/")
```
rcaloras: Modified your change via my comment below then found this issue.

I think there's an edge case here where a trap contains output. If I set a simple debug trap with an echo it's pulling in the output that is produced from the debug trap here. and setting it in the original debug trap function. e.g:

```bash
$ trap
trap -- '' SIGTSTP
trap -- '' SIGTTIN
trap -- '' SIGTTOU

$ trap 'echo "from trap"' DEBUG

$ source bash-preexec.sh 
from trap
from trap
from trap
from trap
from trap
from trap
from trap
from trap
from trap

$ type __bp_original_debug_trap
from trap
__bp_original_debug_trap is a function
__bp_original_debug_trap () 
{ 
    from trap;
    from trap;
    echo "from trap"
}
```

This causes __bp_original_debug_trap to error with the output 'from trap' in the function.

bash-preexec.sh
dimo414: Oof, good catch. I realize my test is also missing the point, since the `DEBUG` trap is invoked immediately. I've got a better test pending.

This would also have been caught if we didn't [silently skip missing functions](https://github.com/rcaloras/bash-preexec/blob/b6a9b9a/bash-preexec.sh#L98) - I'd been thinking about suggesting changing that, I'll file a bug unless you'd prefer to leave it be.

bash-preexec.sh
dimo414: That's really interesting. I think this is due to `set -T`, I'll experiment some more but I may have to go back to something like your `__bp_trap_install_string` solution if `set -T` won't work. I'll get back to you.

rcaloras: sounds good.

bash-preexec.sh
rcaloras: I think skipping missing functions is more pragmatic and is the actual behavior by preexec/precmd in zsh. I think we should be trying to emulate that experience.

bash-preexec.sh
dimo414: Ok, this is fixed. It seems `set -T` is a non-starter, since it causes the DEBUG trap to be run from within command substitutions:

```
$ trap 'echo "from trap"' DEBUG
$ trap_cmd=$(trap -p DEBUG)
from trap
$ set -T
from trap
$ trap_t_cmd=$(trap -p DEBUG)
from trap
$ trap DEBUG
from trap
$ :
$ echo "$trap_cmd"
trap -- 'echo "from trap"' DEBUG
$ echo "$trap_t_cmd"
from trap
trap -- 'echo "from trap"' DEBUG
```

But the pattern I'm using now (capture the trap and clear it in `PROMPT_COMMAND`, then install the new trap in a function) appears to work.

rcaloras:
:+1: :hamburger: :fries: 

@dimo414 :100: :100: :+1: Thanks so much for your contributions. This is great!

dimo414:
Happy to help :)

Issue #52, opened 2017-09-08
Preserved DEBUG traps should be invoked every time, not just on preexec


dimo414:
A coworker pointed out that that the fix to #39 doesn't properly preserve the behavior of the DEBUG trap, because preexec [only triggers on interactive prompts](https://github.com/rcaloras/bash-preexec/blob/a1bab0c/bash-preexec.sh#L133), whereas the prior operation was triggered on every DEBUG trap.

I think this can be fixed by simply moving the old DEBUG trap _out_ of `preexec_functions` and instead `eval`-ing it at the beginning of `__bp_preexec_invoke_exec` - before the short-circuiting checks are made.

The only problem with this is it makes it difficult for someone to clean up the old trap - it's no longer in the trap itself _nor_ in the `preexec` / `preexec_functions`, but some other variable. It's simple enough to document that (e.g. "The old DEBUG trap will be preserved in a `bp_prior_debug_trap` variable") but that widens the API surface and might be abused.

I'll prepare a patch, but I wanted to get some opinions on the approach first.

rcaloras:
@dimo414 interesting observation. Since the original implementation simply overwrote the existing `DEBUG` trap I think anything is really an improvement here :)

I'm in favor of having a variable preserving the contents of the prior debug trap and invoked how you spec'd. Although it widens the API it's definitely a better experience. It does make things a little more complicated, but if we want to support legit preservation it's worth the compromise :+1:

rcaloras:
Assigning over! As always thanks for the contributions :100: :1st_place_medal:

PR #54, opened 2017-09-16
Update DEBUG preserving behavior to fire on all DEBUG traps, not just preexec events

dimo414:
Fixes issue #52.

### DIFF
--- a/bash-preexec.sh
+++ b/bash-preexec.sh
@@ -26,13 +26,13 @@
 #
 #       preexec_functions+=(my_preexec_function)
 #
-#  3. If you have anything that's using the Debug Trap, change it to use
-#     preexec. (Optional) change anything using PROMPT_COMMAND to now use
-#     precmd instead.
+#  3. Consider changing anything using the DEBUG trap or PROMPT_COMMAND
+#     to use preexec and precmd instead. Preexisting usages will be
+#     preserved, but doing so manually may be less surprising.
 #
-#  Note: This module requires two bash features which you must not otherwise be
-#  using: the "DEBUG" trap, and the "PROMPT_COMMAND" variable. prexec_and_precmd_install
-#  will override these and if you override one or the other this will most likely break.
+#  Note: This module requires two Bash features which you must not otherwise be
+#  using: the "DEBUG" trap, and the "PROMPT_COMMAND" variable. If you override
+#  either of these after bash-preexec has been installed it will most likely break.
 
 # Avoid duplicate inclusion
 if [[ "$__bp_imported" == "defined" ]]; then
@@ -44,10 +44,6 @@ __bp_imported="defined"
 # functions, should they want it.
 __bp_last_ret_value="$?"
 __bp_last_argument_prev_command="$_"
-
-# Command to set our preexec trap. It's invoked once via
-# PROMPT_COMMAND and then removed.
-__bp_trap_install_string="trap '__bp_preexec_invoke_exec \"\$_\"' DEBUG;"
 
 # Remove ignorespace and or replace ignoreboth from HISTCONTROL
 # so we can accurately invoke preexec with a command from our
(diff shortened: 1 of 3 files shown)


rcaloras:
@dimo414 

Tried running this locally and looks like we'll invoke the original function more often than originally intended. Is this an issue? Wondering if this is simply a side effect of bash-preexec in general. What do you think?

```bash
elementz@Kashmir:~/git/bash-preexec (dimo414-master)$ trap
trap -- '' SIGTSTP
trap -- '' SIGTTIN
trap -- '' SIGTTOU

elementz@Kashmir:~/git/bash-preexec (dimo414-master)$ trap 'echo "From trap"' DEBUG

elementz@Kashmir:~/git/bash-preexec (dimo414-master)$ ls
From trap
bash-preexec.sh  LICENSE.md  README.md  test  test.sh

elementz@Kashmir:~/git/bash-preexec (dimo414-master)$ pwd
From trap
/home/elementz/git/bash-preexec

elementz@Kashmir:~/git/bash-preexec (dimo414-master)$ source bash-preexec.sh 
From trap
From trap
From trap
From trap
From trap
From trap
From trap
From trap
From trap
From trap
From trap

# Notice the additional invocations here
elementz@Kashmir:~/git/bash-preexec (dimo414-master)$ pwd
From trap
/home/elementz/git/bash-preexec
From trap
From trap

elementz@Kashmir:~/git/bash-preexec (dimo414-master)$ echo
From trap

From trap
From trap
```

dimo414:
I think this is an expected side-effect - I can replicate the same thing without my change by manually tweaking the DEBUG trap:

```
bash-4.3$ source bash-preexec.sh
bash-4.3$ trap -p DEBUG
trap -- '__bp_preexec_invoke_exec "$_"' DEBUG
bash-4.3$ trap '__bp_preexec_invoke_exec "$_"; echo "From trap"' DEBUG
From trap
From trap
bash-4.3$ echo
From trap

From trap
From trap
```

One of the extra traps is triggered by the `PROMPT_COMMAND` invocation, so that's definitely expected. I've looked at the `set -x` output but I don't immediately see why the third trap event is triggered. Regardless, anyone defining a DEBUG trap would need to account for these potential other triggers (as bash-preexec is doing).

rcaloras:
Thanks for following up @dimo414. I think this preserves the integrity of the trap, but I'm not sure this is the best experience. If we weren't manipulating the `DEBUG` trap, do you think this is more or less the expected behavior by comparison to simply invoking on preexec? 

Put another way, do you think invoking `DEBUG` multiple times since we're piggy backing on the trap is better than invoking it on preexec? The experience invoking on preexec could be considered closer to the functionality you'd experience without bash-preexec.

@dimo414 Do you have any examples of `DEBUG` trap usage that we'd want to invoke multiple times in this case other than to simply preserve integrity of the trap? I'm slightly inclined to think that users who would still need to preserve true `DEBUG` functionality would simply augment the trap either by editing directly or editing bash-preexec locally to add it.

dimo414:
Good question - I agree that most people using the DEBUG trap in an interactive session probably *want* preexec semantics and have to jump through hoops to get it (since they're not using this library). However, we can't know that's what was intended. It's true that installing bash-preexec potentially changes the number of times the original DEBUG trap would be triggered, but anyone implementing their own DEBUG trap has to account for the possibility of the trap firing at other times than just preexec, or their implementation is simply wrong.

> The experience invoking on preexec could be considered closer to the functionality you'd experience without bash-preexec.

I don't know if I agree with that - in particular the DEBUG trap fires twice for `$ foo; bar` but preexec only fires once:

```
bash-4.3$ trap 'From trap' DEBUG
bash-4.3$ echo foo; echo bar
From trap
foo
From trap
bar
bash-4.3$ source bash-preexec.sh
From trap
From trap
From trap
bash-4.3$ echo foo; echo bar
From trap
foo
bar
```

I don't have a concrete use-case for needing this distinction or otherwise needing to preserve the DEBUG semantics at the moment, but it does seem a bit odd to me not to do so. That said, maybe it's better to hold off on this fix for now?

rhansen:
I have a DEBUG trap that sets the title of my terminal window to `$BASH_COMMAND` (among other things).  For this usage, multiple invocations of the trap is important:  If I run `foo && bar`, and both `foo` and `bar` are long-running commands, I want the title to display `bar` once `foo` has completed.

akinomyoga:
Why has this been closed? Was it intentional? I still expect the behavior of existing DEBUG traps, which were supposed to be here.

Issue #58, opened 2017-10-05
Incompatible prompt_command settings


jombooth:
Some users of a project I work on that's based on Bash-Preexec modify their prompt command in ~/.bashrc after sourcing the script. In some cases, we've found they'll do this:

<source bash-preexec>
PROMPT_COMMAND+="history -a"

This breaks BP, because prompt_command ends in "__bp_install", so on first execution, the user gets a
"no such command: __bp_installhistory" message.

To fix this particular issue, I suggest wrapping the prompt_command installed by bash_preexec in newlines, ie
__bp_preexec_foo_bar; __bp_install \n

so that appending to PROMPT_COMMAND won't cause problems. Newlines are in general the most universally compatible way of splitting commands in Bash, so we could consider using them wherever a problem might arise, ie,

\n __bp_preexec_foo_bar \n __bp_install \n

This way, even if someone does a PROMPT_COMMAND="echo good morning! $PROMPT_COMMAND", after sourcing bash-preexec, the result will still be correct.

w.r.t. testing this - it would be ideal to test various possible orderings of setting up custom prompt commands and sourcing Bash-Preexec, for example,

adjust PROMPT_COMMAND
source bash_preexec
adjust PROMPT_COMMAND again

source bash_preexec
adjust PROMPT_COMMAND

rcaloras:
Yeah, there used to be semicolons, but we decided to remove them. https://github.com/rcaloras/bash-preexec/pull/41

Newlines sound more forgiving :) want to submit the PR?

rcaloras:
@jombooth closing this guy since I merged the PR. Hoping this is fixed now on master.

PR #59, opened 2017-10-17
Add newlines to PROMPT_COMMAND install string

rcaloras:
- Addresses #58

### DIFF
--- a/bash-preexec.sh
+++ b/bash-preexec.sh
@@ -281,7 +281,7 @@ __bp_install_after_session_init() {
     # trap. __bp_install sets PROMPT_COMMAND to its final value, so these are only
     # invoked once.
     # It's necessary to clear any existing DEBUG trap in order to set it from the install function.
-    PROMPT_COMMAND='__bp_trap_string=$(trap -p DEBUG); trap DEBUG; __bp_install'
+    PROMPT_COMMAND=$'\n__bp_trap_string=$(trap -p DEBUG); trap DEBUG; __bp_install\n'
 }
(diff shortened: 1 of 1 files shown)


rcaloras:
@brandonweeks @jombooth @dimo414 mind reviewing this PR?

bash-preexec.sh
```diff
@@ -281,7 +281,7 @@ __bp_install_after_session_init() {
     # trap. __bp_install sets PROMPT_COMMAND to its final value, so these are only
     # invoked once.
     # It's necessary to clear any existing DEBUG trap in order to set it from the install function.
-    PROMPT_COMMAND='__bp_trap_string=$(trap -p DEBUG); trap DEBUG; __bp_install'
+    PROMPT_COMMAND=$'\n__bp_trap_string=$(trap -p DEBUG); trap DEBUG; __bp_install\n'
```
jombooth: Overall, lgtm; I verified that $' ' quoting should work in the expected way. I think it would be better to replace all the ';'s with newlines, however.

jombooth: Also - it might make sense to double-quote the process substitution:

...__bp_trap_string="$(trap -p DEBUG)"...

rcaloras: Adding the quotes. I feel like the semicolons are appropriate considering it's one line of logic here.

rcaloras: Reconsidering :/  I'll change to all newlines. That's an easy convention to adhere to.

PR #60, opened 2017-10-19
Remove saving prior trap in PROMPT_COMMAND with subshell

rcaloras:
- This was causing crashes in cases where users were doing:
  export PROMPT_COMMAND="history -a; $PROMPT_COMMAND;"

- Original description from contributor:
  $(trap -p DEBUG)" is launching a sub-shell, bash itself. So when people do
  "export PROMPT_COMMAND="history -a; $PROMPT_COMMAND;"", they export form the
  parent shell to subsequent sub-shells. And that creates the infinite loop,
  that eventually segfault bash.

### DIFF
--- a/bash-preexec.sh
+++ b/bash-preexec.sh
@@ -277,12 +277,15 @@ __bp_install_after_session_init() {
       precmd_functions+=(__bp_original_prompt_command)
     fi
 
+    # Save anything already defined in our trap
+    __bp_trap_string="$(trap -p DEBUG)"
+
     # Installation is finalized in PROMPT_COMMAND, which allows us to override the DEBUG
     # trap. __bp_install sets PROMPT_COMMAND to its final value, so these are only
     # invoked once.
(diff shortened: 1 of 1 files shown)


jombooth:
LGTM

rcaloras:
@jombooth :+1:

dimo414:
I'm surprised that moving `__bp_trap_string="$(trap -p DEBUG)"` out of the `PROMPT_COMMAND` doesn't break trap preservation. Did you verify this is still working? It's possible the existing tests don't cover this sort of refactoring well enough.

rcaloras:
@dimo414 You're right that this has broken trap preservation, sorry to have disrupted this feature. I wanted to put this through to remove the bug from our thread with Loic. If you have a way to still preserve it, please submit a PR :+1:

dimo414:
That's fine, the patch is just confusing since it leaves bits of the trap-preservation logic in place (e.g. there's no point setting `__bp_trap_string` at all).

I'll file a separate bug to try to restore it.

PR #62, opened 2017-10-24
Update bash-preexec.sh

lguelorget:
- Avoids infinite recursion with things like export PROMPT_COMMAND="history -a; $PROMPT_COMMAND"
- Preserve trap DEBUG again: revert https://github.com/rcaloras/bash-preexec/pull/60

### DIFF
--- a/bash-preexec.sh
+++ b/bash-preexec.sh
@@ -149,7 +149,7 @@ __bp_preexec_invoke_exec() {
 
     # Don't invoke preexecs if we are inside of another preexec.
     if (( __bp_inside_preexec > 0 )); then
-      return;
+      return
     fi
     local __bp_inside_preexec=1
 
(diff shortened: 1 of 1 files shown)


bash-preexec.sh
```diff
@@ -83,10 +86,17 @@ __bp_interactive_mode() {
 # This function is installed as part of the PROMPT_COMMAND.
 # It will invoke any functions defined in the precmd_functions array.
 __bp_precmd_invoke_cmd() {
-
     # Save the returned value from our last command
```
jombooth: Since this is problem is quite unintuitive, I think putting a comment like

    # Save the returned value from our last command. Note: this MUST be the
    # first thing done in this function.

would help avoid it resurfacing in the future.

Issue #64, opened 2018-01-06
Setting `IFS=_` breaks __bp_precmd_invoke_cmd


gnachman:
I received the following issue report:

https://gitlab.com/gnachman/iterm2/issues/6398

The internals of `__bp_precmd_invoke_cmd` assume `IFS` has its default value. I think you could do this:

1. At the top of `__bp_precmd_invoke_cmd`, save the current value of `IFS` to a local variable
2. Set `IFS` to ` \t\n`
3. At the end of `__bp_precmd_invoke_cmd` restore `IFS` to the saved value.

What do you think?

rcaloras:
@gnachman Thanks for reporting! Did as you suggested storing it. I created a couple small helpers based on https://unix.stackexchange.com/a/264947.  Give #65 a quick look when you get a sec. I'll cut a version as soon as it's merged.

rcaloras:
cc @D630 @dimo414 @lguelorget bash-preexec's ⭐️ contributors.

dimo414:
I worry the proposed fix would essentially introduce the opposite problem, since the user's function would be invoked with bash-preexec's `IFS`. Sure, we could do the same resetting/restoring logic inside the loops too, but that's pretty nasty.

What exactly is the behavior that's IFS-dependent currently? At a glance I don't see it.

rcaloras:
@dimo414 That's true. IFS will be set to bash-preexec's definition for precmd and preexec.

Maybe I'm over complicating things, but if I don't have default IFS our for loops will break. For read statements it's possible to set IFS on the [same line](https://github.com/rcaloras/bash-preexec/commit/6529e23da9400849cc837f47a1bb2a26fe5fc803#diff-9f42bfd9f684d2763a4d12cc481506a1R207). Wonder if I can do that with our loops as well.

rcaloras:
Updated to set original IFS across `precmd` and `preexec` functions

dimo414:
> if I don't have default IFS our for loops will break

Can you share a particular example? I believe that array-splitting (e.g. `for e in "${my_array[@]}"`) is not impacted by IFS, though I certainly could be wrong about that. I wasn't able to break anything with a few attempts.

D630:
Ja, I don't think setting/resetting of IFS is necessary. But you have to take care about (at least):

1. the read builtin. Set IFS as keyword parameter in front of it. Or don't use read in a pipe at all. E.g.:

```bash
# stupid

[...]

PR #65, opened 2018-01-07
Store and unset IFS for arrays and read

rcaloras:
- Fix for #64
- bash-preexec would break if IFS was changed. It should be default for
  the way we're using arrays and read.


rcaloras:
Replaced by #66 for the latest PR.

PR #66, opened 2018-01-15
Fix #64 by quoting functions and set IFS on read

### LINKED ISSUE #64: Setting `IFS=_` breaks __bp_precmd_invoke_cmd

rcaloras:
- Fixes bug where IFS in outer environment could be set to something that
  would cause our inner functions to break on varibale expansion.

- Added a couple unit tests.

### DIFF
--- a/bash-preexec.sh
+++ b/bash-preexec.sh
@@ -106,7 +106,7 @@ __bp_precmd_invoke_cmd() {
         # Test existence of functions with: declare -[Ff]
         if type -t "$precmd_function" 1>/dev/null; then
             __bp_set_ret_value "$__bp_last_ret_value" "$__bp_last_argument_prev_command"
-            $precmd_function
+            "$precmd_function"
         fi
     done
 }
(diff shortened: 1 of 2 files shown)


bash-preexec.sh
```diff
@@ -106,7 +106,7 @@ __bp_precmd_invoke_cmd() {
         # Test existence of functions with: declare -[Ff]
         if type -t "$precmd_function" 1>/dev/null; then
             __bp_set_ret_value "$__bp_last_ret_value" "$__bp_last_argument_prev_command"
-            $precmd_function
+            "$precmd_function"
```
dimo414: It would be good to leave a comment in the file somewhere noting why we need to quote function names.

rcaloras: https://github.com/rcaloras/bash-preexec/commit/331fa90ba5575369f27afbd17ca1afb23e0ce4c2

PR #73, opened 2018-09-12
Detect variables bash-preexec requires writable access to that have been marked readonly

### LINKED ISSUE #72: Detect readonly HISTTIMEFORMAT
I received an issue report from a user who had a read-only HISTTIMEFORMAT variable. It caused an error to be printed on every command executed. The issue report is here:

https://gitlab.com/gnachman/iterm2/issues/7028

You can reproduce it by doing

```
readonly HISTTIMEFORMAT
```

after loading `bash-preexec`, and you get this output on every command:

```
bash: HISTTIMEFORMAT: readonly variable
```

I'm not sure if bash lets you detect that the variable is readonly. I suppose the simplest fix would be to redirect the output of the assignment to /dev/null.

dimo414:
Resolves #72.

### DIFF
--- a/test/include-test.bats
+++ b/test/include-test.bats
@@ -12,3 +12,9 @@
   [[ -n $(type -t __bp_install) ]]
 }
 
+@test "bp should stop installation if HISTTIMEFORMAT is readonly" {
+  readonly HISTTIMEFORMAT
+  run source "${BATS_TEST_DIRNAME}/../bash-preexec.sh"
+  [[ $status != 0 ]]
+  [[ "$output" =~ "HISTTIMEFORMAT" ]]
+}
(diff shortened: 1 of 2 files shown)


test/include-test.bats
```diff
@@ -12,3 +12,7 @@
   [[ -n $(type -t __bp_install) ]]
 }
 
+@test "bp should stop installation if HISTTIMEFORMAT is readonly" {
+  readonly HISTTIMEFORMAT
+  ! source "${BATS_TEST_DIRNAME}/../bash-preexec.sh"
```
rcaloras: Could you add assert the logging is working as well? e.g.
```
  [[ "$output" == "bash-preexec requires write access to HISTTIMEFORMAT" ]]
```

dimo414: Done.

rcaloras:
@dimo414 thanks for the PR! Much appreciated! :+1: :+1: :+1:

Issue #78, opened 2018-11-29
Whitespace is trimmed from commands


steinarvk:
Whitespace is trimmed from commands. For instance, the command " foo   " is recorded as "foo".

This whitespace is stripped when reading the line from the history with the "read" builtin.

Issue #77, opened 2018-11-29
commands with multiline strings are truncated to first line


steinarvk:
Commands that contain multi-line strings are truncated to the first line. For instance, the following command

echo 'foo
bar'

results in this_command="echo 'foo" being passed to the preexec hooks. The "bar" part is entirely lost. This is because "read" only reads a single line.

Issue #17, opened 2015-10-24
Remove subshells and external commands (= sed)


D630:
Hi,

I have opened four pull requests. Don't know, whether they are compatible with Bash on BSD

#13, #14, #15, #16

rcaloras:
Hey @D630!

Thanks for the contributions, I'll give them a look over the next day and pull them in. What would you say are the major benefits of removing the subshells and sed? Simplicity, portability, speed? 

Appreciate it :+1:

D630:
Hey!

It is unnecessary to use sed, because bash can do the job itself (and in the end even faster). Command substitutions are only necessary, if we do need the output of a command. But in #15 and #16 it is sufficient to work with exit/return status. See also [SubShell](http://mywiki.wooledge.org/SubShell).

So, the major benefits are simplicity and speed

rcaloras:
Thank you for the contributions @D630! Much appreciated. :+1: 

Would love your feedback or input on my project [Bashhub](https://github.com/rcaloras/bashhub-client) which uses bash-preexec if you're interested.

PR #79, opened 2018-11-29
Fix issues with capturing certain commands

### LINKED ISSUE #76: $this_command is interpreted by `echo`
A very small number of useless commands are ignored by bash-preexec, because they're [passed to `echo`](https://github.com/rcaloras/bash-preexec/blob/dd689aed44b7485f064256312fa2a420b415ad20/bash-preexec.sh#L202). E.g., the command line `-n` causes bash-preexec to run `echo -n` which does not print `-n`. The solution I've seen is to use `printf '%s\n' ...` instead of `echo`.

steinarvk:
This uses a sed-based regex instead of the read builtin to drop the initial number output by the 'history' command.

The problems with the read builtin are:
  - It only reads a single line.
  - It strips whitespace.

Changing to this sed expression (which removes the initial number from the first line only, assuming that the format of the history line is "%5d%c %s%s\n" as in history.def in the bash source code) resolves both of these issues, fixing #77 and #78 .

This also removes the need for either an echo or a printf, and so also incidentally resolves #76 . (As pointed out on that issue, this could also have been fixed without switching to sed.)

Adds regression tests for the fixed issues. The test code is also adjusted to not have the same echo/printf issue.

### DIFF
--- a/bash-preexec.sh
+++ b/bash-preexec.sh
@@ -199,7 +199,7 @@ __bp_preexec_invoke_exec() {
     fi
 
     local this_command
-    this_command=$(HISTTIMEFORMAT= builtin history 1 | { IFS=" " read -r _ this_command; echo "$this_command"; })
+    this_command=$(HISTTIMEFORMAT= builtin history 1 | sed -E '1 s/^ *[0-9]+[* ] //')
 
     # Sanity check to make sure we have something to invoke our function with.
     if [[ -z "$this_command" ]]; then
--- a/test/bash-preexec.bats
+++ b/test/bash-preexec.bats
@@ -17,7 +17,7 @@ test_echo() {
 }
 
 test_preexec_echo() {
-  echo "$1"
+  printf "%s\n" "$1"
 }
 
 @test "__bp_install_after_session_init should exit with 1 if we're not using bash" {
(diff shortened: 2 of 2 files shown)


dimo414:
I'll note this `read` behavior was introduced in 4822db8f10 (sugested by @d630, who might have some opinions). While I appreciate the goal of [avoiding shelling out to `sed`](https://github.com/rcaloras/bash-preexec/issues/17#issuecomment-151497813) I think in this case it's necessary.

I did experiment with syntax like `[[ $this_command =~ [0-9]+[*\ ]\ (.*)$  ]] && echo "${BASH_REMATCH[1]}"` but at this point (particularly the multi-line issue) I don't think it's worth trying to wrestle with Bash any further.

dimo414:
That seems workable from a quick test.

rcaloras:
@steinarvk thanks for the contributions! :tada: Much appreciated :pray: 

As the current implementation was by @d630, curious your thoughts. :+1:  on this change?

steinarvk:
d630's solution is clever. I like the lack of dependency on the exact format of the bash string. I tried it out in a terminal earlier and it seemed like it would work, but when I paste it in verbatim it causes tests to fail (all three new tests plus two old ones).

I'm sure that's possible to fix. I'd be fine with any implementation as long as it passes the tests. I'd like to get the tests in to prevent regressions, though. If avoiding sed is preferred, feel free to fork from my fork and make a separate PR with my tests + a different implementation.

dimo414:
Can you share the failures you're seeing? You could just commit the change to this branch so we can see the Travis failures.

d630:
So, here is the problem: before bash 4.4, getting the history entry like shown above needs HISTTIMEFORMAT to be set before we append the entry to the list:

```sh
#!/usr/bin/env bats

setup() {
  PROMPT_COMMAND=''        # in case the invoking shell has set this
  HISTTIMEFORMAT=%s
  history -s fake command  # preexec requires there be some history
  __bp_delay_install="true"
  source "${BATS_TEST_DIRNAME}/../bash-preexec.sh"
}
```

In bash 4.4 and above I would use:

```sh
    this_command=$(HISTTIMEFORMAT=%%; builtin history 1)
    this_command=${this_command#*%}
```

rcaloras:
Merging. @steinarvk thanks again for the PR! :+1: :hamburger: :fries:

Issue #80, opened 2018-11-29
bash-preexec fails if user has "nounset" option set


steinarvk:
Some users set the bash option "nounset", which makes evaluating "${foo}" an error if there is no variable named "foo". When expanding as a string, "${foo:-}" can be used to explicitly indicate that it is okay for the variable to be unset and use an empty string instead.

This option currently makes bash-preexec fail all over the place, whether set -u is set before or after it is sourced. Unset variables are referenced many times during initialization and hook execution.

Each individual line is mostly easy to fix, but it's a pain to maintain. We cannot set "nounset" in bash-preexec.sh itself, because that would pollute the user's shell options in a potentially very annoying way. However, if we can configure the bats tests to run under "set -u", we should be able to maintain compatibility with this option.

(Preliminary testing indicates that bats 0.4.0 does not support set -u, but bats 1.1.0 from the bats-core fork does. We may have to change the way we install bats on travis for the necessary changes to the test environment to pass.)

It would be nice to also test "set -x", but even bats 1.1.0 does not seem to support that.

dimo414:
Personally, I don't think supporting `-u` is worth the effort. If you can make it work great, but I think using it (in your shell) is generally an anti-pattern.

+1 for swapping to the bats-core fork though.

steinarvk:
It may be an anti-pattern, but I think it'd be best not to make that choice on behalf of the user. Shells are highly personal things and the less we make them break in mysterious ways for our users, the better, even if they've made strange choices to get there.

The main practical issue is when code that breaks "nounset" is reached from the debug trap that bash-preexec installs, which means the user can't even safely use nounset in contexts where it to all appearances is neatly scoped to their own code. However, it's easier to just make the entire script nounset-safe rather than to try to reason about what is reachable from where.

Supporting it is easy, the only difficulty is supporting it consistently. And given that this happens to be a bash script with a test suite, even that shouldn't be hard. I have the necessary code and test changes in a local client -- will try to put together a PR tomorrow.

rcaloras:
@steinarvk Thanks for the PR and all the reasoning + logic behind it! On first glance, I agree with @dimo414 that setting `-u` in your shell is an anti-pattern and question if it's worth trying to maintain it in bash-preexec. Looking at the code, I feel it also lowers the readability by inserting `{*:-}` in tons of places. 

[...]

PR #82, opened 2018-11-30
Compatibility with "nounset" bash option

steinarvk:
As noted in #80 , the main actual issue here is that because we install a DEBUG trap, our code may run even in contexts where the user thinks they've set bash options for their own code only. As such, it would be nice if we were very well behaved.

It's easier to just run the tests entirely under this option -- that saves us from having to think about which code needs to be nounset-safe and which code does not.

The actual code changes are pretty trivial, it's just a matter of applying them consistently, which the tests will enforce.

This also updates the bats version used for Travis to 1.1.0, cloning from the bats-core github instead of installing from a PPA. The old version of bats couldn't handle running with nounset. The .travis.yml changes are untested, we should see if they work soon enough.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: bash
 before_install:
   # To install bats and test our shell/bash functions
   - git clone -b "v1.1.0" "https://github.com/bats-core/bats-core.git"
-  - sudo ./bats-core/install.sh
+  - sudo ./bats-core/install.sh /usr/local
   - rm -rf ./bats-core
   - sudo apt-get install -qq zsh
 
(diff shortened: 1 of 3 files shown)


steinarvk:
Finally figured it out. There's different behaviour with nounset and array expansion on bash 4.3 and bash 4.4 (which I'm running locally). I initially tried running _bats_ from bash 4.3 and still saw nothing different, but of course what one really has to do is to spawn bats with the path set so that _it_ will execute bash 4.3.

I'll discard this PR and set up a prettier one based off the other PR that just upgrades bats.

PR #83, opened 2018-11-30
upgrade Travis bats version to bats-core 1.1.0

steinarvk:
This is necessary for testing with 'nounset'.

Trying again independently of the other PR which also changes the tests, so as to see which change is actually breaking things.


rcaloras:
:+1: :hamburger: :fries:

PR #85, opened 2018-12-01
Save $PIPESTATUS for precmd and preexec functions to use.

### LINKED ISSUE #81: Copy $PIPESTATUS for precmd_functions.
Would you mind making a copy of [`$PIPESTATUS`](https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html#index-PIPESTATUS) at the same time as you copy `$?`? I don't know of any way to restore PIPESTATUS for each precmd_function, so just having a copy available would be fine.

dseomn:
Unfortunately, I couldn't figure out how to restore it for each
function, so I just made a copy in $BP_PIPESTATUS for those functions to
use.

For documentation of PIPESTATUS, see
https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html#index-PIPESTATUS

Closes #81.

### DIFF
--- a/bash-preexec.sh
+++ b/bash-preexec.sh
@@ -39,8 +39,12 @@ fi
 __bp_imported="defined"
 
 # Should be available to each precmd and preexec
-# functions, should they want it.
+# functions, should they want it. $? and $_ are available as $? and $_, but
+# $PIPESTATUS is available only in a copy, $BP_PIPESTATUS.
+# TODO: Figure out how to restore PIPESTATUS before each precmd or preexec
+# function.
 __bp_last_ret_value="$?"
+BP_PIPESTATUS=("${PIPESTATUS[@]}")
 __bp_last_argument_prev_command="$_"
 
 __bp_inside_precmd=0
@@ -95,9 +99,9 @@ __bp_interactive_mode() {
 # This function is installed as part of the PROMPT_COMMAND.
 # It will invoke any functions defined in the precmd_functions array.
 __bp_precmd_invoke_cmd() {
-    # Save the returned value from our last command. Note: this MUST be the
-    # first thing done in this function.
-    __bp_last_ret_value="$?"
+    # Save the returned value from our last command, and from each process in
+    # its pipeline. Note: this MUST be the first thing done in this function.
+    __bp_last_ret_value="$?" BP_PIPESTATUS=("${PIPESTATUS[@]}")
 
     # Don't invoke precmds if we are inside an execution of an "original
     # prompt command" by another precmd execution loop. This avoids infinite
--- a/test/bash-preexec.bats
+++ b/test/bash-preexec.bats
@@ -108,6 +108,22 @@ test_preexec_echo() {
     [[ "$output" == "251" ]]
 }
 
+@test "precmd should set \$BP_PIPESTATUS to the previous \$PIPESTATUS" {
+  echo_pipestatus() {
+    echo "${BP_PIPESTATUS[*]}"
+  }
+  # Helper function is necessary because Bats' run doesn't preserve $PIPESTATUS
+  set_pipestatus_and_run_precmd() {
+    false | true
+    __bp_precmd_invoke_cmd
+  }
+
+  precmd_functions+=(echo_pipestatus)
+  run 'set_pipestatus_and_run_precmd'
+  [[ $status == 0 ]]
+  [[ "$output" == "1 0" ]]
+}
+
 @test "precmd should set \$_ to be the previous last arg" {
     echo_last_arg() {
       echo "$_"


bash-preexec.sh
```diff
@@ -39,8 +39,12 @@ fi
 __bp_imported="defined"
 
 # Should be available to each precmd and preexec
-# functions, should they want it.
+# functions, should they want it. $? and $_ are available as $? and $_, but
+# $PIPESTATUS is available only in a copy, $BP_PIPESTATUS.
+# TODO: Figure out how to restore PIPESTATUS before each precmd or preexec
+# function.
 __bp_last_ret_value="$?"
+BP_PIPESTATUS=("${PIPESTATUS[@]}")
```
dimo414: I'd suggest `__bp_pipestatus` instead of `BP_PIPESTATUS` to be consistent with other "private" variables bash-preexec uses.

dseomn: I chose `BP_PIPESTATUS` to make this explicitly public, since I couldn't think of another way to pass it to pre{exec,cmd} functions.

dimo414: Ah sorry, I didn't realize this was intentional. If it's part of the public API it should probably be documented (e.g. in the README).

bash-preexec.sh
```diff
@@ -95,9 +99,9 @@ __bp_interactive_mode() {
 # This function is installed as part of the PROMPT_COMMAND.
 # It will invoke any functions defined in the precmd_functions array.
 __bp_precmd_invoke_cmd() {
-    # Save the returned value from our last command. Note: this MUST be the
-    # first thing done in this function.
-    __bp_last_ret_value="$?"
+    # Save the returned value from our last command, and from each process in
+    # its pipeline. Note: this MUST be the first thing done in this function.
+    __bp_last_ret_value="$?" BP_PIPESTATUS=("${PIPESTATUS[@]}")
```
dimo414: By the looks of it `BP_PIPESTATUS` will not be usable from `preexec` because `PIPESTATUS` isn't updated at the time the `DEBUG` trap fires, so it should probably be unset at the end of `__bp_precmd_invoke_cmd` so a stale value isn't propagated.

dimo414:
I played around a bit with trying to preserve `PIPESTATUS` itself, but haven't gotten anything working yet. @dseomn, care to share what you've tried?

One approach would be to set `PIPESTATUS` by invoking a command pipeline that triggers the same exit codes, e.g.:

```shell
__bp_return() { return ${1:?}; }
...
__bp_pipestatus=("${PIPESTATUS[@]}")
...
# somehow invoke...
__bp_return ${PIPESTATUS[0]} | __bp_return ${PIPESTATUS[1]} | ...
```

but I haven't figured out a practical way to do this. I can construct the string with:

```shell
$(ps=("${PIPESTATUS[@]}"); IFS='|'; echo "${ps[*]/#/__bp_return }")
```

But I'm not sure how to execute it - [`eval` doesn't propagate the `PIPESTATUS`](https://stackoverflow.com/q/21878286/113632) of the last pipeline it executes.

---

Needless to say, I'm worried this is pretty brittle behavior, and I wonder if it's really worth adding. I can imagine it getting messy in the future.

bash-preexec.sh
dseomn: I'm not too familiar with the `DEBUG` trap, but wouldn't the first one after the user hits enter expect to see the same values of `$?` and `$[BP_]PIPESTATUS` as immediately before the user hit enter?

dseomn:
> I played around a bit with trying to preserve `PIPESTATUS` itself, but haven't gotten anything working yet. @dseomn, care to share what you've tried?

I also tried building a string and passing it to `eval`. The only other thing I tried (which also didn't work) was something like this:

```bash
f() {
  local ret="$1"; shift
  if [[ $# -gt 0 ]]; then
    f "$ret" | f "$@"
  else
    return "$ret"
  fi
}
...
f "${BP_PIPESTATUS[@]}"
```

I eventually gave up when I realized that the last value in `PIPESTATUS` is not always equal to `$?`. Because of that, I don't think it's possible to preserve both variables at the same time, and I think preserving `$?` is more important.

```
$ false | true
$ ^C
$ printf '%s\n%s\n' "$?" "${PIPESTATUS[*]}"
130
1 0
```

> Needless to say, I'm worried this is pretty brittle behavior, and I wonder if it's really worth adding. I can imagine it getting messy in the future.

Are you saying that restoring `PIPESTATUS` is brittle, or making a copy available in `BP_PIPESTATUS`? I'm not sure I see how making the copy would get messy, other than requiring people who want to use it to use a different variable than they're used to.

rcaloras:
Catching up on this thread. @dseomn does this change still make sense and want it merged?

dseomn:
> Catching up on this thread. @dseomn does this change still make sense and want it merged?

If you're ok with exporting a public variable `BP_PIPESTATUS`, then yes, getting this merged would be great!

rcaloras:
@dseomn thanks for your contributions :+1: :hamburger: :fries:

PR #86, opened 2018-12-06
Remove spurious global variable declarations for __bp_last_ret_value and __bp_last_ret_value

dimo414:
Added in 7e75a3ea and d056f7dc, it doesn't look like they've ever worked (or been used).

Also remove initial declarations for `__bp_inside_precmd` and
`__bp_inside_preexec`, which do not need to be globals.

### DIFF
--- a/bash-preexec.sh
+++ b/bash-preexec.sh
@@ -38,14 +38,6 @@ if [[ "$__bp_imported" == "defined" ]]; then
 fi
 __bp_imported="defined"
 
-# Should be available to each precmd and preexec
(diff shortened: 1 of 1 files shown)


rcaloras:
Closing this and leaving these global declarations. Not positive if they're necessary for the last arguments or ret values, but I did consciously add it in https://github.com/rcaloras/bash-preexec/commit/7e75a3ea4ae41bfc92adce27cadebe25a194072d. I also don't think it's bad practice to have these declared during the outset of our script as they're namespaced and should be safe.

PR #90, opened 2019-02-10
Miscellaneous fixes for sed.

dseomn:
1. Avoid using `\+` because it's not available on some BSDs.
2. Set the locale to avoid number formatting issues.


rcaloras:
@dseomn thanks for the PR! 👍 🍔 🍟

PR #96, opened 2019-07-16
Honor HISTCONTROL "ignorespace" and "ignoreboth"

illia-bobyr:
After 3458480385f81124a9fd9a38ff384e43dd815b1c

    Remove ignorespace from $HISTCONTROL

and after 7e55ac15fee86cdf8e08e966d8b8c9dd2bf7ad03

    Follow up commit for issue #6

    -Replace ignoreboth with simpley ignoredups

this script would remove 'ignorespace' and would replace 'ignoreboth'
with 'ignoredups'.  This effectively disables the functionality of not
adding space prefixed commands into history.  It used to happen
siliently and could be quite confusing to users who use this feature.

This script relies on the command to be in the history, but we can
mostly fix the issue by "manually" removing a whitespace prefixed command
from the history after reading it from there.

### DIFF
--- a/bash-preexec.sh
+++ b/bash-preexec.sh
@@ -86,18 +86,29 @@ __bp_require_not_readonly() {
     done
 }
 
-# Remove ignorespace and or replace ignoreboth from HISTCONTROL
-# so we can accurately invoke preexec with a command from our
-# history even if it starts with a space.
+# Replace "ignorespace" and/or "ignoreboth" in HISTCONTROL with
+# "bash-preexec_ignorespace" and "ignoredups:bash-preexec_ignorespace"
+# respectively.  This is necessary in order for the preexec hook to be able to
+# read the last command from the history, even if it starts with a space.  We
+# then remove commands that start with a space from the history in
+# __bp_load_this_command_from_history, if "bash-preexec_ignorespace" is part of
+# HISTCONTROL.
 __bp_adjust_histcontrol() {
     local histcontrol
-    histcontrol="${HISTCONTROL:-}"
-    histcontrol="${histcontrol//ignorespace}"
-    # Replace ignoreboth with ignoredups
-    if [[ "$histcontrol" == *"ignoreboth"* ]]; then
-        histcontrol="ignoredups:${histcontrol//ignoreboth}"
-    fi
-    export HISTCONTROL="$histcontrol"
+    histcontrol=${HISTCONTROL:-}
+
+    # We might need to delete some of the elements.  It is easier when each
+    # elements has it's own set of separators.
+    histcontrol=":${histcontrol//:/::}:"
+
+    histcontrol=${histcontrol//:ignorespace:/:bash-preexec_ignorespace:}
+    histcontrol=${histcontrol//:ignoreboth:/:ignoredups::bash-preexec_ignorespace:}
+
+    histcontrol=${histcontrol//::/:}
+    histcontrol=${histcontrol#:}
+    histcontrol=${histcontrol%:}
+
+    export HISTCONTROL=$histcontrol
 }
 
 # This variable describes whether we are currently in "interactive mode";
@@ -310,9 +321,69 @@ __bp_in_prompt_command() {
     return 1
 }
 
+# Loads the last executed command from the command history, assiging it into the
+# $this_command variable.
+#
+# If $HISTCONTROL contains "ignorespace", $this_command will be set to an empty
+# string.
+#
+# If $HISTCONTROL contains "bash-preexec_ignorespace", removes the extacted
+# command from the command history, if the command starts with a space.  This
+# simulates the built-in "ignorespace" behavior.
 __bp_load_this_command_from_history() {
-    this_command=$(LC_ALL=C HISTTIMEFORMAT='' builtin history 1)
-    this_command="${this_command#*[[:digit:]][* ] }"
+    local histcontrol
+    histcontrol=${HISTCONTROL:-}
+
+    local ignorespace=''
+
+    # If ignorespace is in $HISTCONTROL, we do not know if the last command in
+    # the history is actually the last executed command.  But it is up to the 
+    if [[ "$histcontrol" =~ (^|:)ignorespace(:|$) \
+        || "$histcontrol" =~ (^|:)ignoreboth(:|$) ]];
+    then
+        if [[ -z "${BP_EMPTY_LAST_COMMAND_ACK:-}" ]]; then
+            cat <<EOM >&2
+bash-preexec: WARNING: HISTCONTROL contains "ignorespace", or "ignoreboth"
+    making it impossible to reliably detect the last executed command.
+    bash-preexec will invoke hook functions with an empty command.
+
+    If this is intentional, set BP_EMPTY_LAST_COMMAND_ACK to avoid this warning:
+
+export BP_EMPTY_LAST_COMMAND_ACK=yes
+
+    You can also replace "ignorespace" with "bash-preexec_ignorespace", to
+    workaround this restriction.  See "\`HISTCONTROL\` interaction" in the
+    bash-preexec documentation.
+EOM
+        fi
+
+        export BP_EMPTY_LAST_COMMAND_ACK=yes
+
+        this_command=''
+        return 0
+    elif [[ "$histcontrol" =~ (^|:)bash-preexec_ignorespace(:|$) ]]; then
+        ignorespace=yes
+    fi
+
+    local history_last_entry
+    history_last_entry=$(LC_ALL=C HISTTIMEFORMAT='' builtin history 1)
+    this_command="${history_last_entry#*[[:digit:]][* ] }"
+
+    # "bash-preexec_ignorespace" in HISTCONTROL indicates we have removed
+    # "ignorespace" from HISTCONTROL, and we need to remove commands that start
+    # with a space from the history ourselves.
+    #
+    # With bash 5.0 or above, we could have just ran
+    #
+    #   builtin history -d -1
+    #
+    # Negative indices for `-d` are not supported before 5.0, so we compute the
+    # length of the history list explicit, to delete the last entry.
+    if [[ -n "$ignorespace" && "$this_command" == " "* ]]; then
+        if [[ "$history_last_entry" =~ ^[[:blank:]]*([0-9]+) ]]; then
+            builtin history -d "${BASH_REMATCH[1]}"
+        fi
+    fi
 
     # Sanity check to make sure we have something to invoke our function with.
     [[ -n "$this_command" ]]
--- a/test/bash-preexec.bats
+++ b/test/bash-preexec.bats
@@ -587,24 +587,170 @@ set_exit_code_and_run_precmd() {
 
 }
 
-@test "__bp_adjust_histcontrol should remove ignorespace and ignoreboth" {
+@test "__bp_adjust_histcontrol shows nothing when \$HISTCONTROL is untouched" {
+    HISTCONTROL="ignoredups"
+    run '__bp_adjust_histcontrol'
+    [ "$output" == '' ]
+}
 
-    # Should remove ignorespace
+@test "__bp_adjust_histcontrol changes ignorespace to bash-preexec_ignorespace" {
+    # Should not touch `ignoredups`.
+    HISTCONTROL="selection_flag1:ignoredups:selection_flag2:"
+    __bp_adjust_histcontrol
+    [ "$HISTCONTROL" == "selection_flag1:ignoredups:selection_flag2:" ]
+
+    # Should replace ignorespace on its own.
+    HISTCONTROL="ignorespace"
+    __bp_adjust_histcontrol
+    [ "$HISTCONTROL" == "bash-preexec_ignorespace" ]
+
+    # Should replace ignorespace.
     HISTCONTROL="ignorespace:ignoredups:*"
     __bp_adjust_histcontrol
-    [ "$HISTCONTROL" == ":ignoredups:*" ]
+    [ "$HISTCONTROL" == "bash-preexec_ignorespace:ignoredups:*" ]
 
-    # Should remove ignoreboth and replace it with ignoredups
+    # Should replace all ignorespace instances.
+    HISTCONTROL="ignorespace:ignoredups:something_else:ignorespace:more_stuff"
+    __bp_adjust_histcontrol
+    [ "$HISTCONTROL" == "bash-preexec_ignorespace:ignoredups:something_else:bash-preexec_ignorespace:more_stuff" ]
+
+}
+
+@test "__bp_adjust_histcontrol changes ignoreboth to ignoredups:bash-preexec_ignorespace" {
+    # Should replace ignoreboth on its own.
     HISTCONTROL="ignoreboth"
     __bp_adjust_histcontrol
-    [ "$HISTCONTROL" == "ignoredups:" ]
+    [ "$HISTCONTROL" == "ignoredups:bash-preexec_ignorespace" ]
 
-    # Handle a few inputs
-    HISTCONTROL="ignoreboth:ignorespace:some_thing_else"
+    # Should replace ignoreboth.
+    HISTCONTROL="ignoreboth:more_flags"
     __bp_adjust_histcontrol
-    echo "$HISTCONTROL"
-    [ "$HISTCONTROL" == "ignoredups:::some_thing_else" ]
+    [ "$HISTCONTROL" == "ignoredups:bash-preexec_ignorespace:more_flags" ]
 
+    # Should replace all ignoreboth instances.
+    HISTCONTROL="ignoreboth:something_else:ignoreboth:even_more_stuff"
+    __bp_adjust_histcontrol
+    [ "$HISTCONTROL" == "ignoredups:bash-preexec_ignorespace:something_else:ignoredups:bash-preexec_ignorespace:even_more_stuff" ]
+
+    # Should replace all ignoreboth and ignorespace instances.
+    HISTCONTROL="ignoreboth:something_else:ignorespace:more_stuff"
+    __bp_adjust_histcontrol
+    [ "$HISTCONTROL" == "ignoredups:bash-preexec_ignorespace:something_else:bash-preexec_ignorespace:more_stuff" ]
+
+    # Should replace all ignoreboth instances.
+    HISTCONTROL="ignoreboth:ignoredups:something_else:ignoreboth:more_stuff"
+    __bp_adjust_histcontrol
+    [ "$HISTCONTROL" == "ignoredups:bash-preexec_ignorespace:ignoredups:something_else:ignoredups:bash-preexec_ignorespace:more_stuff" ]
+
+}
+
+run_bp_load_this_command_from_history_and_print_it() {
+    local this_command
+    __bp_load_this_command_from_history
+    local status=$?
+    printf '%s\n' "$this_command"
+    return $status
+}
+
+@test "__bp_load_this_command_from_history: HISTCONTROL with no ignore" {
+    local command1="this command is in the history"
+    history -s "$command1"
+
+    run run_bp_load_this_command_from_history_and_print_it
+    [ $status -eq 0 ]
+    [ "$output" == "$command1" ]
+}
+
+run_bp_load_this_command_from_history_and_print_BP_EMPTY_LAST_COMMAND_ACK() {
+    local this_command
+    __bp_load_this_command_from_history
+    local status=$?
+    printf 'BP_EMPTY_LAST_COMMAND_ACK: %s\n' "$BP_EMPTY_LAST_COMMAND_ACK"
+    return $status
+}
+
+@test "__bp_load_this_command_from_history: HISTCONTROL with ignorespace shows a warning" {
+    HISTCONTROL="ignorespace"
+
+    local command1="this command is in the history"
+    history -s "$command1"
+
+    run run_bp_load_this_command_from_history_and_print_BP_EMPTY_LAST_COMMAND_ACK
+    [ $status -eq 0 ]
+    [[ "$output" =~ ^'bash-preexec: WARNING: HISTCONTROL contains "ignorespace"' ]] || return 1
+    [[ "$output" =~ 'BP_EMPTY_LAST_COMMAND_ACK: yes'$ ]] || return 1
+}
+
+run_bp_load_this_command_from_history_twice() {
+    __bp_load_this_command_from_history && __bp_load_this_command_from_history
+}
+
+@test "__bp_load_this_command_from_history: HISTCONTROL with ignorespace shows only one warning" {
+    HISTCONTROL="ignorespace"
+
+    local command1="this command is in the history"
+    history -s "$command1"
+
+    run run_bp_load_this_command_from_history_twice
+    [ $status -eq 0 ]
+    [[ "$output" =~ ^'bash-preexec: WARNING: ' ]] || return 1
+    [[ ! ( "$output" =~ ^'bash-preexec: WARNING: '.*'bash-preexec: WARNING' ) ]] || return 1
+}
+
+@test "__bp_load_this_command_from_history: HISTCONTROL with ignorespace shows no warning if acked" {
+    BP_EMPTY_LAST_COMMAND_ACK=yes
+    HISTCONTROL="ignorespace"
+
+    local command1="this command is in the history"
+    history -s "$command1"
+
+    run __bp_load_this_command_from_history
+    [ $status -eq 0 ]
+    [ "$output" == '' ]
+}
+
+run_bp_load_this_command_from_history_and_print_it_and_the_last_history_entry() {
+    local this_command
+    __bp_load_this_command_from_history
+    local status=$?
+
+    local history_last
+    history_last=$(LC_ALL=C HISTTIMEFORMAT='' builtin history 1)
+    history_last="${history_last#*[[:digit:]][* ] }"
+
+    printf '%s\n' "$this_command"
+    printf '%s\n' "$history_last"
+    return $status
+}
+
+@test "__bp_load_this_command_from_history: HISTCONTROL with bash-preexec_ignorespace" {
+    HISTCONTROL="bash-preexec_ignorespace"
+
+    local command1="this command is in the history"
+    history -s "$command1"
+
+    run run_bp_load_this_command_from_history_and_print_it_and_the_last_history_entry
+    [ $status -eq 0 ]
+    [ "$output" == "$command1"$'\n'"$command1" ]
+
+    local command2=" this should not be in the history"
+    history -s "$command2"
+
+    run run_bp_load_this_command_from_history_and_print_it_and_the_last_history_entry
+    [ $status -eq 0 ]
+    [ "$output" == "$command2"$'\n'"$command1" ]
+}
+
+@test "__bp_load_this_command_from_history: HISTCONTROL with ignorespace" {
+    BP_EMPTY_LAST_COMMAND_ACK=yes
+    HISTCONTROL="ignorespace"
+
+    local command1="this command is in the history"
+    history -s "$command1"
+
+    run run_bp_load_this_command_from_history_and_print_it
+    [ $status -eq 0 ]
+    [ "$output" == "" ]
 }
 
 @test "preexec should respect HISTTIMEFORMAT" {
@@ -647,4 +793,46 @@ a multiline string'" ]
     run '__bp_preexec_invoke_exec'
     [ $status -eq 0 ]
     [ "$output" == '-n' ]
+}
+
+run_bp_preexec_invoke_exec_print_last_history_entry() {
+    __bp_preexec_invoke_exec
+    local status=$?
+
+    local history_last
+    history_last=$(LC_ALL=C HISTTIMEFORMAT='' builtin history 1)
+    history_last="${history_last#*[[:digit:]][* ] }"
+
+    printf '%s\n' "$history_last"
+    return $status
+}
+
+@test "When HISTCONTROL contains ignorespace the ignorespace functionality is honoured" {
+    # This test covers functionality that is mostly already covered by
+    # __bp_adjust_histcontrol and __bp_load_this_command_from_history tests
+    # above.  But it represents a more complete, almost an end-to-end scenario,
+    # making sure that small components interact correctly as a whole.
+
+    preexec_functions+=(test_preexec_echo)
+    HISTCONTROL=ignorespace:ignoreboth
+
+    __bp_adjust_histcontrol
+
+    [ "$HISTCONTROL" == "bash-preexec_ignorespace:ignoredups:bash-preexec_ignorespace" ]
+
+    __bp_interactive_mode
+
+    command1="this command is in the history"
+
+    history -s "$command1"
+    run run_bp_preexec_invoke_exec_print_last_history_entry
+    [ $status == 0 ]
+    [ "$output" == "$command1"$'\n'"$command1" ]
+
+    command2=" this should not be in the history"
+
+    history -s "$command2"
+    run run_bp_preexec_invoke_exec_print_last_history_entry
+    [ $status == 0 ]
+    [ "$output" == "$command2"$'\n'"$command1" ]
 }
--- a/README.md
+++ b/README.md
@@ -30,6 +30,8 @@ curl https://raw.githubusercontent.com/rcaloras/bash-preexec/master/bash-preexec
 echo '[[ -f ~/.bash-preexec.sh ]] && source ~/.bash-preexec.sh' >> ~/.bashrc
 ```
 
+NOTE: this script may change your `HISTCONTROL` value by replacing `ignorespace` with `bash-preexec_ignorespace` or replacing `ignoreboth` with `ignoredups:bash-preexec_ignorespace`.  See [`HISTCONTROL` interaction](#histcontrol-interaction) for details.
+
 ## Usage
 Two functions **preexec** and **precmd** can now be defined and they'll be automatically invoked by bash-preexec if they exist.
 
@@ -90,6 +92,35 @@ bash-preexec does not support invoking preexec() for subshells by default. It mu
 export __bp_enable_subshells="true"
 ```
 This is disabled by default due to buggy situations related to to `functrace` and Bash's `DEBUG trap`. See [Issue #25](https://github.com/rcaloras/bash-preexec/issues/25)
+
+## `HISTCONTROL` interaction
+
+In order to be able to provide the last command text to the `preexec` hook, this
+script uses the command history. It reads the last command from the list of the
+executed commands.  If your `HISTCONTROL` contains `ignorespace` (or
+`ignoreboth`), commands that start with a space are not added into the command
+history. When the pre-exec hook is invoked, it can not tell if the last value
+read from the command history is actually the command executed, or if the last
+executed command was hidden, and the command history contains an older command.
+
+To solve this problem, when bash-preexec is loaded, it will check for
+`ignorespace` and `ignoreboth` values in the `HISTCONTROL` variable and replace
+them with `bash-preexec_ignorespace` and `ignoredups:bash-preexec_ignorespace`,
+respectively. It will also show a note once, that this substitution has been
+performed, unless you also set `BP_HISTCONTROL_ACK` to a non-empty value.
+
+When the preexec hook is invoked, there are 3 possibilities now:
+1. `HISTCONTROL` contains `ignorespace` or `ignoreboth`. In this case
+   bash-preexec can not reliably determine the last executed command. The hook
+   will show a notice once, and will be executed with an empty last command.
+   You can avoid the notice completely if you set `BP_EMPTY_LAST_COMMAND_ACK`.
+2. `HISTCONTROL` contains `bash-preexec_ignorespace`. In this case the hook will
+   read the last command from the command history and will remove it from the
+   history if it is prefixed with a whitespace. The hook will be executed with
+   the command text, even if it is a whitespace prefixed command.
+3. `HISTCONTROL` does not contain `ignorespace`, `ignoreboth`, or
+   `bash-preexec_ignorespace`. In this case the hook will read the last command
+   from the command history and run the hook with the last command.
 
 ## Library authors
 If you want to detect bash-preexec in your library (for example, to add hooks to `preexec_functions` when available), use the Bash variable `bash_preexec_imported`:


bash-preexec.sh
```diff
@@ -213,6 +224,18 @@ __bp_preexec_invoke_exec() {
...
+    # setup, we need to remove commands that start with a space from the
+    # history ourselves.
+
+    # Unfortunately, we need bash 5.0 or above, as "history -d" does not
+    # support negative indices in earlier versions and HISTCMD seems to be
+    # unreliable.  At least in the unit tests HISTCMD is just set to 1, after
+    # any number of "history -s" calls.
+    if [[ "${BASH_VERSINFO:-0}" -ge 5 && -n "$__bp_ignorespace" && "$this_command" == " "* ]]; then
+      builtin history -d -1
+    fi
```
cornfeedhobo, line 237: IMHO, this is better done as

```bash
    if [[ -n "$__bp_ignorespace" ]] && [[ "$this_command" == " "* ]]; then
        history -d $(HISTTIMEFORMAT= builtin history 1 | sed '1 s/^ *([0-9][0-9]*)[* ] //')
    fi
```

gaelicWizard:
I've just submitted a very slight improvement to this PR, which uses @cornfeedhobo's suggestion to not require Bash 5+. I couldn't find a way to "fork" a PR, so I just submitted a new one which includes @ilya-bobyr's patch. 

It seems like this unexpected behavior has been brought up numerous times in several PRs and Issues, and I'm highly interested in getting this regression fixed to support all use cases and committed.

ilya-bobyr:
> 
> 
> I've just submitted a very slight improvement to this PR, which uses @cornfeedhobo's suggestion to not require Bash 5+. I couldn't find a way to "fork" a PR, so I just submitted a new one which includes @ilya-bobyr's patch.
> 
> It seems like this unexpected behavior has been brought up numerous times in several PRs and Issues, and I'm highly interested in getting this regression fixed to support all use cases and committed.

Awesome :)
Sorry, I did not incorporate @cornfeedhobo's suggestion - I did not see any reaction from the repo owner.
So was a bit reluctant to make any efforts myself %)

gaelicWizard:
Thanks @ilya-bobyr! Could you alsö rebase on master? It can't be merged as is 👍

tessus:
@rcaloras @dimo414 may I ask, why this hasn't been merged (except that it needs a rebase)? I believe that ignoring `ignoreboth` is a serious issue when people trust that important commands are not logged.

Does this PR has any side-effects that we might not be aware of? Any chance we can discuss it?

akinomyoga:
There is an alternative suggestion in #119, so we anyway need to decide which one to pick.

> Does this PR have any side-effects that we might not be aware of?

- The user cannot change the setting after the session starts.
- The effects are emulated, but the values are still removed from `HISTCONTROL`, which might confuse users and other frameworks. (Note: this is also mentioned in the PR change to README)
- `HISTIGNORE` is not considered.

But maybe it's still better than nothing.

bash-preexec.sh
```diff
@@ -70,15 +71,25 @@ __bp_require_not_readonly() {
...
-# history even if it starts with a space.
+# Remove ignorespace and or replace ignoreboth from HISTCONTROL so we can
+# accurately invoke preexec with a command from our history even if it starts
+# with a space.  We then remove commands that start with a space from the
+# history "manually", if either "ignorespace" or "ignoreboth" was part of
+# HISTCONTROL.
 __bp_adjust_histcontrol() {
     local histcontrol
-    histcontrol="${HISTCONTROL//ignorespace}"
+    if [[ "$HISTCONTROL" == *"ignorespace"* || "$HISTCONTROL" == *"ignoreboth"* ]]; then
```
akinomyoga, line 81:
```suggestion
    if [[ ":$HISTCONTROL:" == *":ignorespace:"* || ":$HISTCONTROL:" == *":ignoreboth:"* ]]; then
```

bash-preexec.sh
```diff
@@ -70,15 +71,25 @@ __bp_require_not_readonly() {
...
+      __bp_ignorespace=yes
+    fi
+    histcontrol="${HISTCONTROL//ignorespace:}"
+    histcontrol="${histcontrol//:ignorespace}"
+    histcontrol="${histcontrol//ignorespace}"
```
akinomyoga, line 86: We can perform adjustments of `ignoreboth` and `ignoredups` at once.

```suggestion
    histcontrol=:${histcontrol//:/::}:
    histcontrol=${histcontrol//:ignorespace:}
    histcontrol=${histcontrol//:ignoreboth:/:ignoredups:}
    histcontrol=${histcontrol//::/:}
    histcontrol=${histcontrol#:}
    histcontrol=${histcontrol%:}
```

bash-preexec.sh
```diff
@@ -239,6 +250,23 @@ __bp_preexec_invoke_exec() {
         return
     fi
 
+    # If we have removed "ignorespace" or "ignoreboth" from HISTCONTROL
+    # during setup, we need to remove commands that start with a space from
+    # the history ourselves.
+
+    # With bash 5.0 or above, we could have just ran
```
akinomyoga, line 257: We can switch the implementation based on `BASH_VERSINFO`. Reducing a fork for Bash 5.0 is still nice.

akinomyoga (approved):
I prefer the #96 approach to #119. `BASH_COMMAND` is not reliable. After either one is merged, I'll submit a PR of further changes addressing the points in https://github.com/rcaloras/bash-preexec/pull/96#issuecomment-1950298227.

@rcaloras Could you take a look at this?

akinomyoga:
@ilya-bobyr Would it be possible to take a bit of time to resolve the conflicts?

bash-preexec.sh
ilya-bobyr: Applied.

bash-preexec.sh
ilya-bobyr: Applied.

ilya-bobyr:
Rebased and applied 2 out of 3 suggestions.
Let me know what you want me to do with the last one.

bash-preexec.sh
tessus: While I agree with you partly, I have to say that keeping compatibility with ancient bash versions is also a problem. Why do people have to "suffer", because of it?
How many people are really on bash <5?

I come from a performance background and doing a regex where it is not needed seems a waste to me. Of course this is not a huge strain on resources, but preexec is a hook, which means it is called all the time, thus is should be as sleek and performant as possible, wouldn't you agree?

Maintenance is a good point, however, I would still favor the code path that is most likely taken, which in my book is the bash >=5 path.

akinomyoga: Thank you for updating. Bash-preexec already has a version check and supports Bash 3.1+. I'm not a maintainer, but I think it is not realistic to drop the support for Bash < 5. The situation for Bash and bash-preexec seems to be different from a typical project.

- Maybe we can drop the support for 3.1, but I think we still need to support 3.2 because 3.2 will continue to be shipped with macOS for the license issue.
- Also, LTS distributions will continue to provide a Bash version for ten years. Since Bash is tightly bound to the system, it's non-trivial to update the system-installed Bash for distributors. Although users can install the latest version of Bash in their home directory, it is difficult to force all the users to do so correctly.
- In a typical project, another option might be to maintain several branches of versions, where new features are added to the well-tested latest version, while older versions receive the corrections for real problems. However, the *feature set* of bash-preexec is stable and no feature was added for a long time, yet the implementation has been always unstable and incomplete. There hasn't been any *stable* version, and we would probably want to apply many patches retroactively even if we adopted such a model. Anyway, this model would add even more maintenance costs.

Although you seem to suggest two options, dropping support for Bash < 5.0 entirely or `if`-statement, if you care about the test coverage, I think we will need to keep the current version based on the old Bash features.

tessus: I've updated my comment above. I just wanted to clarify that I didn't mean to drop support for bash < 5. I meant the `if` statement should be used to allow bash 5 users to use the more performant processing option.

akinomyoga (approved):
Anyway, either the current version or the `if`-statement works for me. No need to further delay it. Thank you for your quick response to the old PR! I'm not a maintainer, but let me approve the PR of the current state.

akinomyoga (approved):
@rcaloras

ilya-bobyr:
I've created a separate PR to update the test runner image: https://github.com/rcaloras/bash-preexec/pull/171
But pulled in that change into this PR, to make sure the tests pass.

akinomyoga:
@ilya-bobyr Thanks for updating the PR!

@rcaloras Thank you for checking the existing PRs! Could you also take a look at this PR? I think the change in this PR should be incorporated, though this PR has been rebased many times to be consistent with the master branch. The original manipulation of `HISTCONTROL` in the current `master` has some issues, and this PR tries to *improve* the situation (though it doesn't perfectly solve all possible conflicts with various settings). This PR also adds documentation about the issue.

akinomyoga (approved):

ilya-bobyr:
I've rebased against the latest master.
It removed https://github.com/rcaloras/bash-preexec/pull/171, as it was already merged.

akinomyoga (approved):

akinomyoga (approved):

ilya-bobyr:
It does not seem that I can merge this change even with an approval from @akinomyoga.
GitHub says an approval from someone with write access is still required.

akinomyoga:
In my understanding, the only active person who has write access is @rcaloras. Anyway, people having write access need to merge, so the requirement of reviews from people with write access seems trivial. I just send approvals to the maintainers to express my request to merge this PR.

akinomyoga (approved):
@rcaloras

akinomyoga (approved):
Actually, it was not simply a style issue, but Bats, the underlying testing framework, distinguishes `[ ]` and `[[ ]]`. Bats doesn't treat the latter as a test item.

I recently obtained write access to this repository. I can technically merge this PR and want to, but I haven't yet understood my role in this repository. Let us ping @rcaloras: This PR has been open for seven years, and the author has been consistently responsive during that period. I feel it is unfair to delay merging further. I think the actual changes also make sense.

rcaloras:
Thanks for the ping and continued interest all. On the surface, I've been against this change since we manipulate the history preferences on purpose and intentionally to get a clean version of the command string. It's effectively a compromise/requirement even if some users may be unaware of the manipulation.

What this change does is effectively then try to preserve them even though it'd no longer be configured to do so in `HISTCONTROL` right? `HISTCONTROL` then is no longer a source of truth and instead we're trying to have bash-preexec managae this history feature. I'm not a fan of this as it feels like a compromise on top of a compromise.  Ideally, would love to simply find a better way to not have to manipulate a user's history preferences at all - while still being able to provide the command string for preexec. Not sure if the latest versions of Bash have made it possible. Prior thoughts also here: https://github.com/rcaloras/bash-preexec/pull/117#issuecomment-887851926

With that said, @akinomyoga I'll defer to you if you want to merge this as a primary maintainer now. My only request @illia-bobyr would be to rewrite as well to remove the use of `sed` for perf and convention.

bash-preexec.sh
```diff
@@ -397,6 +406,24 @@ __bp_invoke_preexec_from_ps0() {
...
+    #
+    #   builtin history -d -1
+    #
+    # Negative indices for `-d` are not supported before 5.0, so we compute the
+    # length of the history list explicit, to delete the last entry.
+    if [[ -n "$__bp_ignorespace" && "$this_command" == " "* ]]; then
+        builtin history -d "$(
+            export LC_ALL=C
+            HISTTIMEFORMAT= history 1 | sed '1 s/^ *\([0-9][0-9]*\).*/\1/'
```
rcaloras: Can we do this without sed? Believe it'll be more efficient and we've removed it from all other places.

rcaloras (changes requested):

illia-bobyr:
> Thanks for the ping and continued interest all. On the surface, I've been against this change since we manipulate the history preferences on purpose and intentionally to get a clean version of the command string. It's effectively a compromise/requirement even if some users may be unaware of the manipulation.
> 
> What this change does is effectively then try to preserve them even though it'd no longer be configured to do so in `HISTCONTROL` right? `HISTCONTROL` then is no longer a source of truth and instead we're trying to have bash-preexec managae this history feature. I'm not a fan of this as it feels like a compromise on top of a compromise. Ideally, would love to simply find a better way to not have to manipulate a user's history preferences at all - while still being able to provide the command string for preexec. Not sure if the latest versions of Bash have made it possible.

I can see how the fact that `HISTCONTROL` is not showing the actual behavior can be confusing.
But, it is a choice between two suboptimal solutions.

Without this change, if I use `bash-preexec`, I lose an ability to prevent commands from being added into my history.
It is a security feature that I normally enable by setting the corresponding `HISTCONTROL` value in my `.bashrc`.
At first glance, `bash-preexec` seems to have no relationship to this security feature and is useful in a number of prompt technics.

What happened to me is that I thought that certain commands are not recorded.
As my `HISTCONTROL` had `ignoreboth` added in my `.bashrc`.  And it was working.
Later, I realized that some passwords ended up in my history, despite the setting.
Took me a while to figure out who was responsible.  And the only choice was to remove `bash-preexec`, reducing my prompt usability.

I think a compromise of the `HISTCONTROL` value being honored, but the value not reflecting the initial setting is less surprising compared to the above.

In the first option, the value of `HISTCONTROL` is changed from the one I've selected, and the security sensitive behavior I have selected is gone.
In the second option the value of `HISTCONTROL` is changed from the one I've selected, but the security sensitive behavior I have selected stays in place.
It seems to me that the second option is strictly better.

I doubt that changing `HISTCONTROL` locally or temporarily is a very common behavior.
But I agree that it would be better if there would be something similar to `$BASH_COMMAND`, that would provide the complete pipeline, with no aliases resolved.

I am not aware of a way to get the ideal solution.  `$BASH_COMMAND` is not capturing the whole command line.
If someone finds a better solution, they can send a PR.
Meanwhile, if this PR is accepted, a real frustration is removed for users who actually rely on the `ignorespace` behavior.  What I want is the command not be stored in my history.  How is it done, is, after all, secondary.

> Prior thoughts also here: [#117 (comment)](https://github.com/rcaloras/bash-preexec/pull/117#issuecomment-887851926)
> 
> With that said, @akinomyoga I'll defer to you if you want to merge this as a primary maintainer now. My only request @illia-bobyr would be to rewrite as well to remove the use of `sed` for perf and convention.

Removed `sed` from `bash-preexec.sh` and cleaned the code a bit more.
Also removed `sed` from tests.

---

P.S. Not sure if it is OK to keep `sed` in tests.
I think operating on strings using regular expressions is more common, compared to the "longest prefix / shortest prefix" thinking that one is reduced to when forced to use only bash.
[Here is a comparison of the tests before and after](https://github.com/illia-bobyr/bash-preexec/compare/d2376b77fa4908da5af6154e1cb99a6a5968acdf..8ef77ade140a0c48b348b4c132263fe5c347de71#diff-9452363e390bda524fdadd7ee05e6db87bae6dd4444edfb4c29382d78afa14c5R667).
I've included the latest changes from the upstream as well - only the last 2 chunks of the diff are my changes.

akinomyoga:
@rcaloras Thank you for sharing your thoughts!

> On the surface, I've been against this change since we manipulate the history preferences on purpose and intentionally to get a clean version of the command string. It's effectively a compromise/requirement even if some users may be unaware of the manipulation.

The status of the current `master` branch is that we already modify the user's history configurations on purpose. This PR introduces the explanations in README, but it is just an explanation of the existing compromise/requirement in the `master` branch. This PR rather tries to improve the situation by trying to emulate the original behavior. The addition of the following lines is the only essential change of this PR:

https://github.com/rcaloras/bash-preexec/blob/8ef77ade140a0c48b348b4c132263fe5c347de71/bash-preexec.sh#L425-L432

Or, in the previous version you reviewed,

https://github.com/rcaloras/bash-preexec/blob/d2376b77fa4908da5af6154e1cb99a6a5968acdf/bash-preexec.sh#L420-L425

> What this change does is effectively then try to preserve them even though it'd no longer be configured to do so in HISTCONTROL right? HISTCONTROL then is no longer a source of truth and instead we're trying to have bash-preexec managae this history feature.

I think we can leave a custom name, such as `__bp_ignorespace`, in `HISTCONTROL` so that the behavior can be modified later by updating `HISTCONTROL`. Also, when `HISTCONTROL` contains `__bp_ignorespace`, the user can find that `ignorespace` is now emulated by `bash-preexec.sh`.

> I'm not a fan of this as it feels like a compromise on top of a compromise.

Yes, you are right...

> Ideally, would love to simply find a better way to not have to manipulate a user's history preferences at all - while still being able to provide the command string for preexec. Not sure if the latest versions of Bash have made it possible. Prior thoughts also here: https://github.com/rcaloras/bash-preexec/pull/117#issuecomment-887851926

I don't think it is possible to reliably get the previous command even in the latest Bash version.

- Another way is to completely remove the history manipulation and give up the feature to get the complete command line (as in #117). However, in https://github.com/rcaloras/bash-preexec/pull/117#issuecomment-887851926, you seem to prioritize the feature to get the complete command line over preserving the user's history preferences.
- Or we do not modify `HISTCONTROL`, but we may still try to get the previous command line through `builtin history 1`. The obtained command line becomes wrong for the command line starting with a whitespace, but it works for other cases.

bash-preexec.sh
```diff
@@ -402,6 +411,26 @@ __bp_invoke_preexec_from_ps0() {
...
+    #
+    # Negative indices for `-d` are not supported before 5.0, so we compute the
+    # length of the history list explicit, to delete the last entry.
+    if [[ -n "$__bp_ignorespace" && "$this_command" == " "* ]]; then
+        local last_entry
+        last_entry=$(LC_ALL=C HISTTIMEFORMAT='' builtin history 1)
+        # Strip leading whitespace
+        last_entry="${last_entry#"${last_entry%%[![:space:]]*}"}"
+        # Use just the entry index
+        builtin history -d "${last_entry%%[^0-9]*}"
```
akinomyoga: Couldn't this be simplied as follows?

```suggestion
        if [[ $(LC_ALL=C HISTTIMEFORMAT='' builtin history 1) =~ [0-9]+ ]]; then
            builtin history -d "$BASH_REMATCH"
        fi
```

illia-bobyr: I think it will work, yes.
It feels a bit too loose thought.
What do you think about

```bash
        if [[ "$(LC_ALL=C HISTTIMEFORMAT='' builtin history 1)" =~ ^[[:space:]]*([0-9]+) ]]; then
            builtin history -d "${BASH_REMATCH[1]}"
        fi
```

?

I would personally prefer to split it into two statements, to make a bit more readable as well:

```bash
        local last_entry
        last_entry=$(LC_ALL=C HISTTIMEFORMAT='' builtin history 1)
        if [[ "$last_entry" =~ ^[[:space:]]*([0-9]+) ]]; then
            builtin history -d "${BASH_REMATCH[1]}"
        fi
```

akinomyoga: Either works. No particular preferences from me.

akinomyoga: Could you use `[[:blank:]]` instead of `[[:space:]]`? `[[:space:]]` includes extra control characters such as `\f`, `\r`, `\v`, `\n`.

illia-bobyr: I do not have a strong preference either.
[Updated](https://github.com/rcaloras/bash-preexec/compare/8ef77ade140a0c48b348b4c132263fe5c347de71..2c903c8344e6d94382d013fb57b68b47b566efc8) - please take a look.
Let me know if you want it changed.

bash-preexec.sh
```diff
@@ -402,6 +411,26 @@ __bp_invoke_preexec_from_ps0() {
...
+    # With bash 5.0 or above, we could have just ran
+    #
+    #   builtin history -d -1
+    #
+    # Negative indices for `-d` are not supported before 5.0, so we compute the
+    # length of the history list explicit, to delete the last entry.
+    if [[ -n "$__bp_ignorespace" && "$this_command" == " "* ]]; then
+        local last_entry
+        last_entry=$(LC_ALL=C HISTTIMEFORMAT='' builtin history 1)
+        # Strip leading whitespace and take the entry index
```
akinomyoga: Stripping isn't explicitly performed as before. The stripping is effectively performed by only picking `BASH_REMATCH[1]`, but it's implicit.

```suggestion
        # Take the entry index
```

Or actually can we remove the comment line? It's obvious.

bash-preexec.sh
```diff
@@ -86,18 +87,26 @@ __bp_require_not_readonly() {
...
-    histcontrol="${HISTCONTROL:-}"
-    histcontrol="${histcontrol//ignorespace}"
-    # Replace ignoreboth with ignoredups
-    if [[ "$histcontrol" == *"ignoreboth"* ]]; then
-        histcontrol="ignoredups:${histcontrol//ignoreboth}"
+    histcontrol=${HISTCONTROL:-}
+    histcontrol=":${histcontrol//:/::}:"
+
+    if [[ "$histcontrol" == *":ignorespace:"* || "$histcontrol" == *":ignoreboth:"* ]]; then
+      __bp_ignorespace=yes
```
akinomyoga: > From https://github.com/rcaloras/bash-preexec/pull/96#issuecomment-4218905388
>
> > What this change does is effectively then try to preserve them even though it'd no longer be configured to do so in HISTCONTROL right? HISTCONTROL then is no longer a source of truth and instead we're trying to have bash-preexec managae this history feature.
> 
> I think we can leave a custom name, such as `__bp_ignorespace`, in `HISTCONTROL` so that the behavior can be modified later by updating `HISTCONTROL`. Also, when `HISTCONTROL` contains `__bp_ignorespace`, the user can find that `ignorespace` is now emulated by `bash-preexec.sh`.

I started to think that the above is actually a good idea. I mean, we include the value `__bp_ignorespace` in `HISTCONTROL` instead of storing the information in an independent variable `__bp_ignorespace`. Then, we check `*:__bp_ignorespace:*` before manually performing "`history -d -1`". It should be noted that `__bp_ignorespace` is just ignored by Bash. This way, the user may later remove `__bp_ignorespace` to cancel the effect of `ignorespace`, etc. Maintaining the custom settings for the history recording in `HISTCONTROL` seems the correct way. What do you think?


```suggestion
      histcontrol=${histcontrol//:__bp_ignorespace:}:__bp_ignorespace:
```

**edit**: I modified the above suggestion so that it removes existing entries of `__bp_ignorespace` to avoid duplicates.

bash-preexec.sh
akinomyoga: Thank you. Closing the conversation. There is an additional small comment, but it's discussed in a new thread https://github.com/rcaloras/bash-preexec/pull/96#discussion_r3063651519.

test/bash-preexec.bats
```diff
@@ -648,3 +647,37 @@ a multiline string'" ]
...
+
+    command1="this command is in the history"
+
+    history -s "$command1"
+    run '__bp_preexec_invoke_exec'
+    [ $status == 0 ]
+    [ "$output" == "$command1" ]
+    local history_last
+    history_last=$(LC_ALL=C HISTTIMEFORMAT='' builtin history 1)
+    history_last="${history_last#*[[:digit:]][* ] }"
```
akinomyoga: Could you use `__bp_load_this_command_from_history`? The same for line 678.

illia-bobyr:
Sorry for the delay.
I think the idea of using a custom entry in `$HISTCONTROL` is very interesting, and it will get us closer to the ideal solution.

Here is an updated version that uses `bash-preexec_ignorespace` in `$HISTCONTROL` to signal that the built-in behavior was adjusted and is now handled by `bash-preexec`.

This turned out to be a pretty substantial change.
I've added a bunch of tests to cover all(?) the new logic and edge cases.

Initially I've used `bp_ignorespace`, but then I thought that `bash-preexec_ignorespace` is less cryptic.  Especially considering that, hopefully, most users would be unaware of this functionality.  And the only situation that might care would be when something strange is happening.  In which case an abbreviation would only be a hurdle, during a debugging process.

For the same reason I've added verbose warnings in cases when the behavior could be confusing.
One is shown when `$HISTCONTROL` value is modified.
The other one is shown when `$HISTCONTROL` contains `ignorespace`, despite our efforts to remove it.
Both warnings can be disabled by explicitly setting an "ack" variable.
And are shown only once.

The behavior is described in details in the [`README.md` now](https://github.com/rcaloras/bash-preexec/pull/96/changes#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5).
So, I'm not repeating it here.

Please take a look and let me know what you think.

bash-preexec.sh
illia-bobyr: Removed the comment.

bash-preexec.sh
illia-bobyr: Created a new version based on this idea.

test/bash-preexec.bats
illia-bobyr: New version is different here, but it is possible that a similar argument can be made there.
I still extract values from history explicitly in the test, rather than calling `__bp_load_this_command_from_history`.

It was a deliberate decision in the old version.
With the idea being, we do not want to rely on the code that we are testing while we are testing it.
In the old version `__bp_load_this_command_from_history` was pretty trivial, so it was a somewhat weak argument.

But in the currently proposed version, I've moved quite a lot of functionality into `__bp_load_this_command_from_history`, so, I think, it now is a pretty strong argument for not using `__bp_load_this_command_from_history` as a helper in tests.

I can add a helper function that wraps specifically the "take the last element from history" logic.
It is only repeated twice, so I'm not sure if it is worth the effort :)
But I can do it, if you insist.

bash-preexec.sh
```diff
@@ -86,18 +86,71 @@ __bp_require_not_readonly() {
...
-    export HISTCONTROL="$histcontrol"
+
+    histcontrol=${histcontrol//:ignorespace:}
+    histcontrol=${histcontrol//:ignoreboth:}
+    histcontrol=${histcontrol//:ignoredups:}
+
+    if [[ -n "$ignoredups" ]]; then
+        histcontrol="${histcontrol}:ignoredups:"
+    fi
+    histcontrol="${histcontrol}:bash-preexec_ignorespace:"
```
akinomyoga: The logic seems far more complicated than it should be. The new implementation appears to attempt to resolve the duplicates that the user has introduced, but that's unrelated to what we need. I don't think we should do extra modifications (in particular if it significantly increases lines of code).

```suggestion
    histcontrol=${histcontrol//:ignorespace:/:bash-preexec_ignorespace:}
    histcontrol=${histcontrol//:ignoreboth:/:ignoredups::bash-preexec_ignorespace:}
```

bash-preexec.sh
```diff
@@ -86,18 +86,71 @@ __bp_require_not_readonly() {
...
+
+    histcontrol=${histcontrol//::/:}
+    histcontrol=${histcontrol#:}
+    histcontrol=${histcontrol%:}
+
+    if [[ -z "${BP_HISTCONTROL_ACK:-}" ]]; then
```
akinomyoga: Does this mean we start to show messages always when the user sets `ignorespace` or `ignoredups`? I don't think we should do this. Setting `ignorespace` or `ignoredups` isn't a special setting. In addition, `bash-preexec` has recently been included by the integration code of many terminals by default, and users don't necessarily recognize it. I don't think that always requiring `BP_HISTCONTROL_ACK` when the user wants to set `ignorespace` and `ignoredups` is a good idea. In addition, we haven't required this, while we've modified `HISTCONTROL` for a long time.

Could you make the feature of showing this message opt-in instead of opt-out?

bash-preexec.sh
akinomyoga: The name `bash-preexec_ignorespace` (with different separators `-` and `_`) feels a bit awkward to me. In particular, `_` appears to have a stronger connectivity, so it looks "bash" + "preexec_ignorespace" at first glance. We already have variables in the name `bash_preexec_imported` (introduced by #123), so if you don't like `__bp_`, can we instead use `bash_preexec_ignorespace`?

bash-preexec.sh
```diff
@@ -402,6 +515,7 @@ __bp_invoke_preexec_from_ps0() {
 # this function returns `0`.
 __bp_invoke_preexec_functions() {
     local lastexit=$1 lastarg=$2 this_command=$3
+
```
akinomyoga: This change in the empty line should be canceled now because there are no surrounding changes.

bash-preexec.sh
akinomyoga: Also, I would like to perform the re-adjustment of `HISTCONTROL` in the `preexec` processing.

After the shell initialization, `HISTCONTROL` may be modified by user commands, so we can check and adjust `HISTCONTROL` in the `precmd` timing (which is similar to `postexec`). We are already adjusting `PROMPT_COMMAND` in [`__bp_precmd_invoke_cmd` through `___bp_install_prompt_command`](https://github.com/illia-bobyr/bash-preexec/blob/9c30f75b298143388315d6fa6ecd4aaa38397fad/bash-preexec.sh#L293), so we can also call `__bp_adjust_histcontrol` there. As a related matter, the current PR head tries to detect the user's modification in `__bp_load_this_command_from_history` and doesn't try to resolve the issue of grabbing the wrong entry. In this perspective, I think that should be done in the `precmd` timing, and doing so would give a chance to adjust `HISTCONTROL` before the commands starting with whitespace are lost, which reduces to the idea of re-adjustment of `HISTCONTROL` in the `preexec` processing.

If you want to show messages by default when user commands modify `HISTCONTROL` to include `ignorespace` or `ignoredups`, `__bp_adjust_histcontrol` may issue the messages only when it is called from `__bp_precmd_invoke_cmd` (i.e., when we detect the change of `HISTCONTROL` after the shell initialization completes). The function `__bp_precmd_invoke_cmd` may pass an argument to `__bp_adjust_histcontrol` to enable the message.

bash-preexec.sh
```diff
@@ -86,18 +86,71 @@ __bp_require_not_readonly() {
...
+        cat <<EOM >&2
+bash-preexec: NOTE: HISTCONTROL has been modified to support command extraction.
+    Original: $histcontrol_orig
+    Modified: $histcontrol
+    You can set BP_HISTCONTROL_ACK to a non-emtpy value to avoid this message.
+
+export BP_HISTCONTROL_ACK=yes
+
+EOM
+        export BP_HISTCONTROL_ACK=yes
```
akinomyoga: `export` is unnecessary.

Actually, `export` for `HISTCONTROL` is unnecessary too, but I haven't pointed it out because that's an existing code, and fixing it is unrelated to the scope of this PR.

bash-preexec.sh
akinomyoga: The prefix `BP_` seems unclear to users, equally as `__bp_ignorespace` that you rejected in `HISTCONTROL`. 

Can we instead use `bash_preexec_histcontrol_*` in the same variable naming as `bash_preexec_imported`? `BP_PIPESTATUS` provides the same feature as `PIPESTATUS`, so I feel it is unavoidable to use a similar and short name (i.e., using `bash_preexec_pipestatus` in place of `PIPESTATUS` makes the code appearance significantly be different, and it doesn't associate itself with the original `PIPESTATUS` compared to `BP_PIPESTATUS`). I'm not sure if we should extend the naming `BP_` to more places, while we already have `bash_preexec_imported`. The opposite direction might be to unify `bash_preexec_imported` to `bp_imported` or `BP_IMPORTED`, but we already changed the variable name from `__bp_imported` to `bash_preexec_imported`, and I'd like to avoid changing the variable name multiple times.

bash-preexec.sh
illia-bobyr: Removed.

bash-preexec.sh
illia-bobyr: > The logic seems far more complicated than it should be. The new implementation appears to attempt to resolve the duplicates that the user has introduced, but that's unrelated to what we need. I don't think we should do extra modifications (in particular if it significantly increases lines of code).

I'm not sure, I would agree to the "far too complicated" classification %)
I think it is pretty straightforward.
First we check, if there are any flags in `$HISTCONTROL` that we care about.
Then we act accordingly.
The deduplication is more of a side effect, rather than the goal.
In order to show a warning, I need to know if `$HISTCONTROL` was changed.
And separating the check from the action here helps.

But if you really feel strongly about it, I've replaced explicit checks with an unconditional replacement.

Also, if the warning is not going to be displayed, the need to check if any changes were performed is gone.

---

> The name `bash-preexec_ignorespace` (with different separators `-` and `_`) feels a bit awkward to me. In particular, `_` appears to have a stronger connectivity, so it looks "bash" + "preexec_ignorespace" at first glance. We already have variables in the name `bash_preexec_imported` (introduced by #123), so if you don't like `__bp_`, can we instead use `bash_preexec_ignorespace`?

This is not a name of a variable inside bash-preexect code, but something a user might see without the bash-preexec context.
This is why, I figured, using the proper name of the extension is the best.

I think it is absolutely the case that some people can read it as "bash" + "preexec_ignorespace".
But using the same separate does not make it less so, does it?

The reason I've used a different separator is exactly to make it more likely that people will read it as "bash-preexec" and "ignorespace".
I do not care much, so I would go with whatever you will agree to merge.
But if you ask me, I'm not sure that `bash_preexec_ignorespace` is really better than `bash-preexec_ignorespace`, especially considering the extension is called "bash-preexec".

We may consider is a different separator.
For example, it can be `bash-preexec/ignorespace`, or `bash_preexec/ignorespace`.
I think with a forward slash the chances of someone reading it as "bash" and "preexec/ignorespace" are lower.

What do you think?

---

P.S. It is better to create new threads for unrelated suggestions.
It makes it easier to understand which replies apply to which suggestions, and they can be resolved independently of one another.

bash-preexec.sh
illia-bobyr: Why is it unnecessary?
I do not want to show the same warning again.
If the "ack" flag is set when the user profile is loaded, the warning is displayed once, and all the subshells started by the shell that read `.bashrc` will inherit the ack flag.
If `export` is removed, I think, the warning will be displayed multiple times - by every new bash instance.

bash-preexec.sh
illia-bobyr: > Does this mean we start to show messages always when the user sets `ignorespace` or `ignoredups`? I don't think we should do this. Setting `ignorespace` or `ignoredups` isn't a special setting. In addition, `bash-preexec` has recently been included by the integration code of many terminals by default, and users don't necessarily recognize it. I don't think that always requiring `BP_HISTCONTROL_ACK` when the user wants to set `ignorespace` and `ignoredups` is a good idea. In addition, we haven't required this, while we've modified `HISTCONTROL` for a long time.
> 
> Could you make the feature of showing this message opt-in instead of opt-out?

Yes, the idea was to show the warning once if `$HISTCONTROL` is modified.
Users who use bash-preexec can set the ack flag variable in their `.bashrc` before loading bash-preexec.  And they would not see this warning anymore.
This would be the most explicit way, hopefully making sure nothing unexpected happens due to the `$HISTCONTROL` manipulation.

If bash-preexec is part of an existing profile configuration that the user did not setup themselves, indeed it might be confusing that after an update they now see a warning.
I do not think, it is not such a big deal, as the warning explicitly explains why is it shown and one can just copy/paste a line from the warning into their `.bashrc` to never see it again, right?

Here is the warning text:

```
bash-preexec: NOTE: HISTCONTROL has been modified to support command extraction.
    Original: $histcontrol_orig
    Modified: $histcontrol
    You can set BP_HISTCONTROL_ACK to a non-emtpy value to avoid this message.

export BP_HISTCONTROL_ACK=yes
```

I personally would prefer a warning, but I can see how some people might be upset that their existing profile starts to show warnings after an update.

I have removed it from the new version, as you seems to be against it.
But if my arguments for it sounds reasonable to you, I'll bring it back.

---

> Also, I would like to perform the re-adjustment of `HISTCONTROL` in the `preexec` processing.
> 
> After the shell initialization, `HISTCONTROL` may be modified by user commands, so we can check and adjust `HISTCONTROL` in the `precmd` timing (which is similar to `postexec`). We are already adjusting `PROMPT_COMMAND` in [`__bp_precmd_invoke_cmd` through `___bp_install_prompt_command`](https://github.com/illia-bobyr/bash-preexec/blob/9c30f75b298143388315d6fa6ecd4aaa38397fad/bash-preexec.sh#L293), so we can also call `__bp_adjust_histcontrol` there. As a related matter, the current PR head tries to detect the user's modification in `__bp_load_this_command_from_history` and doesn't try to resolve the issue of grabbing the wrong entry. In this perspective, I think that should be done in the `precmd` timing, and doing so would give a chance to adjust `HISTCONTROL` before the commands starting with whitespace are lost, which reduces to the idea of re-adjustment of `HISTCONTROL` in the `preexec` processing.
> 
> If you want to show messages by default when user commands modify `HISTCONTROL` to include `ignorespace` or `ignoredups`, `__bp_adjust_histcontrol` may issue the messages only when it is called from `__bp_precmd_invoke_cmd` (i.e., when we detect the change of `HISTCONTROL` after the shell initialization completes). The function `__bp_precmd_invoke_cmd` may pass an argument to `__bp_adjust_histcontrol` to enable the message.

If we invoke `__bp_adjust_histcontrol` on every prompt update it would make it impossible for the user to add `ignorespace` or `ignoreboth` into `$HISTCONTROL` at all.
Current behavior is that `$HISTCONTROL` is updated only during bash-preexec setup.
And I personally would prefer to keep it this way.
I think changing that would be orthogonal to the intent of this PR.
So, if you are OK with that, I would like to avoid this change for now.

The second warning I've added covers this case - one that is shown when `$HISTCONTROL` contains `ignorespace` or `ignoreboth` despite the adjustment at installation time.
bash-preexec does not prevent the user from setting `$HISTCONTROL` to anything they want, but informs if the new configuration breaks some bash-preexec functionality.
Not all tools that rely on bash-preexec need the last command text in the first place.
I personally use bash-preexec for the command execution stats.
And it does not care for the last command text.
But I do use `ignorespace`.

---

> The prefix `BP_` seems unclear to users, equally as `__bp_ignorespace` that you rejected in `HISTCONTROL`.
> 
> Can we instead use `bash_preexec_histcontrol_*` in the same variable naming as `bash_preexec_imported`? `BP_PIPESTATUS` provides the same feature as `PIPESTATUS`, so I feel it is unavoidable to use a similar and short name (i.e., using `bash_preexec_pipestatus` in place of `PIPESTATUS` makes the code appearance significantly be different, and it doesn't associate itself with the original `PIPESTATUS` compared to `BP_PIPESTATUS`). I'm not sure if we should extend the naming `BP_` to more places, while we already have `bash_preexec_imported`. The opposite direction might be to unify `bash_preexec_imported` to `bp_imported` or `BP_IMPORTED`, but we already changed the variable name from `__bp_imported` to `bash_preexec_imported`, and I'd like to avoid changing the variable name multiple times.

There is a difference in what is used in `$HISTCONTROL` and a variable user sets in `.bashrc`.

Flags in `$HISTCONTROL` need to make sense in the context of just bash.
And this is why, I think, using an abbreviation could make it harder to understand what is going on.
I imagine people would only care about the value in `$HISTCONTROL` when something does not work as expect.
I think, this was one of the arguments from `@rcaloras` against this PR - that the behavior is different from what `$HISTCONTROl` prescribes.
So, if I, as a user of bash, am confused as to why commands are not retained, I would look at `$HISTCONTROL`.  I might not be aware that bash-preexec is involved in any way.
Seeing `bp_ignorespace` gives me less information on to who is erasing commands from the command history, compared to `bash-preexec_ignorespace`.

`BP_HISTCONTROL_ACK` (and `$BP_EMPTY_LAST_COMMAND_ACK`), on the other hand, are variables that the user will set fully aware that they are dealing with bash-preexec.
They will set these variables either after reading the documentation, or after seeing the warning messages.
And they add assignments into these variables into their `.bashrc`.
Most likely after the code that loads bash-preexec.
I have this in my `.bashrc`:
```bash
if [[ -d "$bashMods/bash-preexec" && -r "$bashMods/bash-preexec/bash-preexec.sh" ]]; then
  source "$bashMods/bash-preexec/bash-preexec.sh"
fi
```

I think it is clear enough what the `BP_` prefix would mean in this context.
And turning it into `BASH_PREXEC_` is a bit of an overkill.

Compare

```bash
if [[ -d "$bashMods/bash-preexec" && -r "$bashMods/bash-preexec/bash-preexec.sh" ]]; then
  export BP_HISTCONTROL_ACK=yes
  source "$bashMods/bash-preexec/bash-preexec.sh"
fi
```

to

```bash
if [[ -d "$bashMods/bash-preexec" && -r "$bashMods/bash-preexec/bash-preexec.sh" ]]; then
  export BASH_PREEXEC_HISTCONTROL_ACK=yes
  source "$bashMods/bash-preexec/bash-preexec.sh"
fi
```

But if you want, I can change it.

I do not think a comparison with `bash_preexec_imported` and `BP_PIPESTATUS` is 100% apples to apples.  Both `$bash_preexec_imported` and `$BP_PIPESTATUS` are used by other libraries that rely on bash-preexec.
While `$BP_HISTCONTROL_ACK` and `$BP_EMPTY_LAST_COMMAND_ACK` are more of an end user interface.
But I can see how one would want to have a consistent interface.

illia-bobyr:
Thank you for a quick review.
Please take a look again.
I've rebased, but GitHub can show [a diff vs the previous version](https://github.com/rcaloras/bash-preexec/compare/9c30f75b298143388315d6fa6ecd4aaa38397fad..7f7cd20be4c5af343b589ee3a2ee51c26fcbb3cb).

tessus:
Is this ready? It's been open for 7 years. It would be really nice for preexec to honor ignorespace.

illia-bobyr:
> Is this ready? It's been open for 7 years. It would be really nice for preexec to honor ignorespace.

I've finished the next version a while back.
It is waiting for a review.

If you are interested, take a look at the diff.
Perhaps, with a review by a third party there will be a slightly higher chance of getting this in?

akinomyoga:
It's not ready. I think we've agreed on the overall direction, but I think small details haven't been concluded. I requested changes in the previous review, but the replies to most of them seemed to be "I don't agree but will follow if you insist even though I don't change my mind". Then, I somehow lost my appetite to continue the discussion because that implies that, no matter what I'd say next, it would be considered coercion. I don't want to say anything more. If the author will blindly follow anything I request, it's probably better to commit changes myself.

akinomyoga:
I'll later take a look at the changes again, then I'll push commits.

illia-bobyr:
> It's not ready. I think we've agreed on the overall direction, but I think small details haven't been concluded. I requested changes in the previous review, but the replies to most of them seemed to be "I don't agree but will follow if you insist even though I don't change my mind". Then, I somehow lost my appetite to continue the discussion because that implies that, no matter what I'd say next, it would be considered coercion. I don't want to say anything more. If the author will blindly follow anything I request, it's probably better to commit changes myself.

I've made changes based on your review.
And if you look at the diff versus the previous version it is not insubstantial: https://github.com/rcaloras/bash-preexec/compare/9c30f75b298143388315d6fa6ecd4aaa38397fad..7f7cd20be4c5af343b589ee3a2ee51c26fcbb3cb

But I do have some disagreements.
So why saying "I'll make a change if you insist" is wrong?

Once case when I said it was with regard to the suggestion on using some functions in the test cases.
I explained why I think my version has merit, but I really do not care too much.

I do not see why your response has to be coercion.
There is an option that you look at my arguments, and reconsider your suggestions.
Or, you may argue against it.
Just like you didn't accept my initial proposal, it seems reasonable to assume I may not accept all of your suggestions either, no?

Especially if you are saying that you think we have agreed on the overall direction, I see no reason why there could not be another round of reviews.

akinomyoga:
> I've made changes based on your review. And if you look at the diff versus the previous version it is not insubstantial: https://github.com/rcaloras/bash-preexec/compare/9c30f75b298143388315d6fa6ecd4aaa38397fad..7f7cd20be4c5af343b589ee3a2ee51c26fcbb3cb

I've noticed that but that is not very much consistent with the argument in https://github.com/rcaloras/bash-preexec/pull/96#discussion_r3114424332, where you told "But if you really feel strongly about it". Although I was about to continue the argument, at the time where you've made the change, I hadn't shown anything about whether I *strongly felt* that. I'm wondering if you've already concluded that I *strongly felt* that.

> But I do have some disagreements.

That's welcome.

> So why saying "I'll make a change if you insist" is wrong?

I'm not sure if that was wrong or not. That depends on your intent of the words. If you wanted to close the discussion, concluding that I'm stubborn and wouldn't hear anything, those words would correctly carry your intent somehow, though there might have been a better word choice. If you are still willing to continue the discussion, I'm afraid to say that the word choices like "if you *really* feel *strongly* about it" or "if you *insist*" aren't very correct. They (particularly, really/strongly/insist) carry different connotations, occasionally also including frustration, regardless of the true intent. Those are the words that typically appear after a long discussion, to stop and conclude the discussion, when it turns out to be hard to form the full agreement. Or there might be a possibility that it is a protocol to tell "I now changed my mind to agree with you" in a certain culture where people are unallowed to change their mind formally (i.e., unallowed to say something inconsistent with their previous opinion) and instead recommended to pretend to reluctantly obey because of the pressure. Actually, it is not so correct in the sense that I didn't know how to respond. To avoid guesswork and interpret it literally, it was reasonable to stop the discussion and tell the final dicision on my discretion, but I wasn't sure.

> Once case when I said it was with regard to the suggestion on using some functions in the test cases. I explained why I think my version has merit, but I really do not care too much.

So you do not care too much?

> I do not see why your response has to be coercion. There is an option that you look at my arguments, and reconsider your suggestions. Or, you may argue against it.

I looked the arguments and chose the option to continue the discussion with counterarguments. You've expanded long arguments, so the reply would be expected to be necessarily long. However, as even my initial requests (which is shorter) *seemed* to make you feel that I'm trying to really strongly insist something, I anticipated that the next longer arguments would be effectively coerce you to follow my *insistence*, which is not my intent.

> Just like you didn't accept my initial proposal, it seems reasonable to assume I may not accept all of your suggestions either, no?

That's reasonable.

> Especially if you are saying that you think we have agreed on the overall direction, I see no reason why there could not be another round of reviews.

So you expect continued discussions?

I'm still not totally sure what would be expected in the next review, the final dicision or the discussion, but I guess the discussion still continues?

illia-bobyr:
[I apologize for a slow reply; I was a bit busy over the last few days]

> > I've made changes based on your review. And if you look at the diff versus the previous version it is not insubstantial: [`9c30f75b298143388315d6fa6ecd4aaa38397fad..7f7cd20be4c5af343b589ee3a2ee51c26fcbb3cb` (compare)](https://github.com/rcaloras/bash-preexec/compare/9c30f75b298143388315d6fa6ecd4aaa38397fad..7f7cd20be4c5af343b589ee3a2ee51c26fcbb3cb)
> 
> I've noticed that but that is not very much consistent with the argument in [(earlier comment)](https://github.com/rcaloras/bash-preexec/pull/96#discussion_r3114424332), where you told "But if you really feel strongly about it". Although I was about to continue the argument, at the time where you've made the change, I hadn't shown anything about whether I _strongly felt_ that. I'm wondering if you've already concluded that I _strongly felt_ that.

It is difficult to infer what people think beyond what is explicitly stated.  So, I'm doing my best to avoid judgment and try to consider all possible options you might have in mind.  This is also the reason I'm stating explicitly when an argument I'm making is not something I care much about.

> > But I do have some disagreements.
> 
> That's welcome.

Glad to hear it.  I also do not want to argue too much.
The way I see it is that even though, so far, I am the one who has spent more time on this PR, going forward it will be the maintainers responsibility.
And as such, the maintainer is the person who would need to decide if they are OK with a particular implementation.

> > So why saying "I'll make a change if you insist" is wrong?
> 
> I'm not sure if that was wrong or not. That depends on your intent of the words. If you wanted to close the discussion, concluding that I'm stubborn and wouldn't hear anything, those words would correctly carry your intent somehow, though there might have been a better word choice. If you are still willing to continue the discussion, I'm afraid to say that the word choices like "if you _really_ feel _strongly_ about it" or "if you _insist_" aren't very correct. They (particularly, really/strongly/insist) carry different connotations, occasionally also including frustration, regardless of the true intent. Those are the words that typically appear after a long discussion, to stop and conclude the discussion, when it turns out to be hard to form the full agreement. Or there might be a possibility that it is a protocol to tell "I now changed my mind to agree with you" in a certain culture where people are unallowed to change their mind formally (i.e., unallowed to say something inconsistent with their previous opinion) and instead recommended to pretend to reluctantly obey because of the pressure. Actually, it is not so correct in the sense that I didn't know how to respond. To avoid guesswork and interpret it literally, it was reasonable to stop the discussion and tell the final dicision on my discretion, but I wasn't sure.

I certainly did not want to say that you are stubborn or make any kind of statements about you personally.  Sorry if you inferred that from what I was saying.
English is my second language, so I might be less tactful than a native speaker.  On top of that, I prefer a slightly more direct communication style, as, hopefully, it leaves less room for misunderstanding.
But, with text-based communication and different cultures, it still happens, as we can see.

In my defense, this is how I talk to my colleagues in the office. In the US, it is totally fine to say, "If you really feel strongly about it, then ...".
I think the maintainer's opinion should have more weight.
So even if I both disagree and think that I have a reasonable argument, I can still accept your position if you feel strongly about it.
In most cases, every option is still a compromise of multiple pros and cons.

As for the "social dances" from different cultures, you are talking about.
My personal opinion would be that it is better to avoid such things as much as possible.
There is a huge diversity of social traditions, and for most open-source contributors, English is a second language.
As such, it would greatly impede both the communication itself and the work progress.
I do not suggest people be rude, of cause.
But by not stating ones intent directly, we only increase the chances of misunderstandings.

> > Once case when I said it was with regard to the suggestion on using some functions in the test cases. I explained why I think my version has merit, but I really do not care too much.
> 
> So you do not care too much?

My focus is mostly to get the feature in.
I do not have a strong preference for how exactly it has to be implemented.
It is obviously less work for me if what I have already done is good enough and is accepted.
But I can also make some changes.  And to a certain degree, I can make a change even if I do not personally agree with it, if it leads to the feature being accepted.

In the latter case, you do not necessarily need to persuade me.  You feeling strongly about it might be enough.

> > I do not see why your response has to be coercion. There is an option that you look at my arguments, and reconsider your suggestions. Or, you may argue against it.
> 
> I looked the arguments and chose the option to continue the discussion with counterarguments. You've expanded long arguments, so the reply would be expected to be necessarily long. However, as even my initial requests (which is shorter) _seemed_ to make you feel that I'm trying to really strongly insist something, I anticipated that the next longer arguments would be effectively coerce you to follow my _insistence_, which is not my intent.
> 
> > Just like you didn't accept my initial proposal, it seems reasonable to assume I may not accept all of your suggestions either, no?
> 
> That's reasonable.
> 
> > Especially if you are saying that you think we have agreed on the overall direction, I see no reason why there could not be another round of reviews.
> 
> So you expect continued discussions?
> 
> I'm still not totally sure what would be expected in the next review, the final dicision or the discussion, but I guess the discussion still continues?

I cannot tell what you think about the updated version without you writing it down.
If you like the latest version enough to include it, then there can be a final decision to include it.
If you think the latest version is not good enough and you have the time/energy to explain what needs to be changed, then please go ahead and do so.
I still have some time budget for this PR.

PR #99, opened 2019-08-28
Change [[ to [ in tests where possible

spl:
* Using `[[ ... ]]` on older versions of Bash (e.g. 3.2 on macOS) causes the test to always pass.
* Includes `test/README.md` that explains the problem and suggest solutions for future tests.

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -1,0 +1,22 @@
+Testing `bash-preexec`
+======================
+
+**Note on test conditions**
+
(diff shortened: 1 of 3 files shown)


spl:
For more information on the problem with `[[ ... ]]` in Bats, see https://github.com/sstephenson/bats/issues/49.

rcaloras:
@spl thank you for your contributions! :hamburger: :fries: :tada: :+1:

PR #101, opened 2019-11-30
fix: variable error when bash set -o nounset

adoyle-h:


rcaloras:
@adoyle-h Thanks for the PR :+1: :hamburger: :fries:

Issue #97, opened 2019-07-28
Could the restriction that bash preexec is the last thing loaded be relaxed?


gnachman:
I occasionally receive issue reports that bash_preexec is clobbering someone's PROMPT_COMMAND. For example: https://gitlab.com/gnachman/iterm2/issues/7961

I think it might be possible to do some string twiddling to figure out of PROMPT_COMMAND got changed between the time bash_preexec was loaded and when __bp_install gets run, but I suspect there's a stronger reason why this is not wise. What's the history here?

rcaloras:
Thanks for opening @gnachman! This is actually a regression as originally bash-preexec preserved `PROMPT_COMMAND` during it's delayed install moving the contents to be invoked during `precmd`. I noticed this a few months back as part of https://github.com/rcaloras/bash-preexec/issues/39#issuecomment-450531807, which I believe was the issue/pr that broke it. This can and should be fixed as having the requirement that bash-preexec being sourced last is super brittle.

rcaloras:
cc @dimo414 who I was hoping could submit the PR for the fix. I'm willing to take a crack, but haven't been an active contributor the past year.

dimo414:
Sorry I missed this ping; @gnachman do you have a fix in mind? I'd be happy to help review a PR.

Conceptually, though, I think the interplay here is always going to be a bit nuanced, and I'm not certain it will ever be possible to make bash-preexec _totally_ transparent to other behavior running in the same shell. Bash's APIs simply aren't robust enough. So it's a question of what sort of use-cases can be better supported without negatively impacting other situations (a balance that #50 may not have properly struck, to be clear).

gnachman:
I don't quite understand the pathology. If you do this after sourcing bash-preexec:
```
PROMPT_COMMAND="$PROMPT_COMMAND; echo hi"
```

then you get:

```
bash: PROMPT_COMMAND: line 5: syntax error near unexpected token `;'
bash: PROMPT_COMMAND: line 5: `; echo hi'
```

I can't for the life of me figure out why. It's not inside any function in bash-preexec.

dimo414:
[...]

PR #110, opened 2020-08-20
Allow bash-preexec to be included at any point

rcaloras:
- Relaxes requirement that bash-prexec.sh needs to be included as last
thing in bash_profile.sh
- Preserves existing prompt command in PROMPT_COMMAND variable. May make
sense to still move this to a function in precmd.
- Should address #97 and #39
- Keeps existing traps preserved as preexec functions
- Removes trailing \n on PROMPT_COMMAND added for #58.


dimo414:
Apologies for not taking a look earlier. The PR looks great!

PR #112, opened 2020-09-16
Reimplement the string-manipulation functions to write to variables instead of stdout

dimo414:
This change allows us to avoid unnecessary command substitutions which are (relatively speaking) quite expensive. Although `__bp_sanitize_string` is only used during installation `__bp_trim_whitespace` is invoked in a loop in the critical path as part of `__bp_preexec_invoke_exec`. Eliminating these subshells will reduce bash-preexec's overhead.

See https://github.com/dimo414/prompt.gem/commit/9764cdb3 for another example of a similar change.

Note: the `printf -v` syntax was added to Bash in v3.1: https://mywiki.wooledge.org/BashFAQ/061

### DIFF
--- a/bash-preexec.sh
+++ b/bash-preexec.sh
@@ -85,24 +85,26 @@ __bp_adjust_histcontrol() {
 # and unset as soon as the trace hook is run.
 __bp_preexec_interactive_mode=""
 
+# Trims leading and trailing whitespace from $2 and writes it to the variable
+# name passed as $1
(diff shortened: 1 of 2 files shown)


rcaloras:
Wow! Playing with a bit locally timing my prompt and its looking like ~2x faster. @dimo414 thank you for this change :pray: :tada: Much appreciated!

dimo414:
Nice! That's even better than I expected. Thanks for the quick review!

Issue #115, opened 2021-02-15
[Bug] HISTCONTROL ignore* setting should be respected


cornfeedhobo:
I understand the current reasoning for removing `ignorespace` and `ignoreboth` but I think they break such an important expectation for users that, if set, preexec should manually cleanup history, to guarantee this expectation is maintained.

rcaloras:
@cornfeedhobo thanks for the feedback! This has been brought up a few times in different ways. Here's a few thoughts related that could help or shed light:

1. There's an open PR currently for what I believe you're suggesting. My take on this is that it inflates the job bash-preexec is doing to account for this edge case. In addition, actual histcontrol settings will be inconsistent with the result since bash-preexec will effectively be managing the 'ignorespace' usecase and overriding histcontrol. 
2. If `$1` isn't needed by `preexec`. You can delete `bp_adjust_histcontrol` invocation in the script and preserve original functionality. (See earlier question on this which is used by iTerm2 https://github.com/rcaloras/bash-preexec/issues/48)
3. Another option is for us to not manipulate histcontrol is to instead use `$BASH_COMMAND` in place of using history when a command is prefixed with a space. The issue with this is it creates other edge cases where it doesn't capture multiple commands on a single line (e.g. `|` or `&&`)

Appreciate any thoughts or other options.

cornfeedhobo:
@rcaloras Thanks for the thoughtful reply! I had not see iTerm's issue when searching. After playing around and thinking about this more, I've come to the conclusion that preexec should do this heavy lifting. I realize that this will make histcontrol inconsistent, but given the fact that this library already requires total control over DEBUG and PROMPT_COMMAND, I think adding this inconsistency is safer than users realizing that commands have been saved to history that contain sensitive information.

I will go through our codebase and ask the team what they think about removing `bp_adjust_histcontrol`.

gaelicWizard:
I've come to understand over the last 24 hours (after I submitted an admittedly grumpy PR to just rip out the history manipulation) that the use case is largely to set the window title or similar with the actual command line entered by the user. Of course this can alsö be used for logging, or perhaps debugging. Aliases are alsö lost from $BASH_COMMAND, so the as-typed may appear substantially different. (E.g., `alias halo='echo $BASH_COMMAND'; halo` prints `echo $BASH_COMMAND` not `halo`.)

[...]

Issue #48, opened 2017-07-14
Question on HISTCONTROL


gnachman:
After adopting bash-preexec for a script I maintain, I received a bug report that it changes `HISTCONTROL`: https://gitlab.com/gnachman/iterm2/issues/5787#note_34771330

Based on the comment in the code, I think I'd rather preserve HISTCONTROL than get leading and trailing spaces correctly reported (I strip them anyway):

```
# Remove ignorespace and or replace ignoreboth from HISTCONTROL
# so we can accurately invoke preexec with a command from our
# history even if it starts with a space.
```

Is the comment accurate? Are there any traps if I remove the call to __bp_adjust_histcontrol? Will anything worse happen than losing some whitespace at the beginning of a command?

I tested it and it seems safe, but I thought I'd check before unleashing this change in prod.

rcaloras:
@gnachman Thanks for opening. I added `bp_adjust_histcontrol` because bash-preexec currently depends on history for getting the entered command to pass to `preexec`. For reference: https://github.com/rcaloras/bash-preexec/blob/master/bash-preexec.sh#L181 If you don't use `$1` in preexec, removing it is perfectly safe. Are you using `$1` in preexec?

This is the most accurate way to get the command, but unfortunately doesn't work in the scenario where someone has `ignoreboth` or `ignorespace`. It won't record the command, but preexec will still be invoked grabbing the last line in history. Alternatively you can get the entered command from `$BASH_COMMAND` but this isn't exactly as the user has typed it (e.g. expanding variables and aliases I believe).

rcaloras:
btw, excited to hear you're using bash-preexec for item2. Love your project. Keep up the great work :+1:

gnachman:
That explains it. I am not using $1 in preexec. You can see the combined code here:

https://iterm2.com/shell_integration/bash

I think I'll remove the call to bp_adjust_histcontrol. If someone has two scripts that use bash-preexec they'll need to do some hacking.

rcaloras:
Awesome, sounds like it should be all good. :+1:

PR #117, opened 2021-07-26
Remove all history manipulation

gaelicWizard:
Remove mangling of HISTCONTROL and remove use of history as a mechanism for detecting the "current" command. Instead, use $BASH_COMMAND like Jesus intended.

### DIFF
--- a/bash-preexec.sh
+++ b/bash-preexec.sh
@@ -80,7 +80,14 @@ __bp_adjust_histcontrol() {
     if [[ "$histcontrol" == *"ignoreboth"* ]]; then
         histcontrol="ignoredups:${histcontrol//ignoreboth}"
     fi;
-    export HISTCONTROL="$histcontrol"
+	
+	if [[ -z "${__bp_suppress_histcontrol_warning:-}" && "$HISTCONTROL" == *ignore+(dups|space)* ]]
+	then
+		echo "bash-preexec is unable to determine full command line when \$HISTCONTROL contains 'ignorespace' (or 'ignoreboth')." >&2
+		echo "Those options have been removed from \$HISTCONTROL." >&2
+	fi
+	
+    HISTCONTROL="$histcontrol"
 }
 
 # This variable describes whether we are currently in "interactive mode";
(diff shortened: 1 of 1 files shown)


dimo414:
Can you detail what, if any, behavior differences users will observe with this change? I assume this would be a breaking change?

rcaloras:
@gaelicWizard thanks for the initiative with the PR and @dimo414 thanks for chiming in. 

My sentiment on this change is the same as mentioned prior in: https://github.com/rcaloras/bash-preexec/issues/115#issuecomment-779395772 I don't believe the regression of not capturing multiple commands is an acceptable one and favor the trade off here with overriding history to get the command line as its typed. 

With that said, @gaelicWizard if you don't require $1 on preexec, feel free to to remove the history manipulation as mentioned https://github.com/rcaloras/bash-preexec/issues/48#issuecomment-315352858.

If there's another way to get access to the command exactly as it's typed without history in Bash I'd love to update this!

gaelicWizard:
Sorry, this PR is marked draft because I both haven't committed everything nor put in any explanation.

After reading other open issues and PRs, I'm going to actually try to just make the history manipulation conditional rather than remove it. It's now clear to me that it's necessary for a particular use case given constraints.

My motivation is twofold: (1) don't change things behind my back; (2) allow hook to run before each command, rather than before each command _line_.

dimo414:
>  allow hook to run before each command, rather than before each command line.

That's totally reasonable to want, but it's not how bash-preexec is implemented (by design). Are you sure you even want to use bash-preexec? Maybe you just want to implement your own DEBUG trap?

gaelicWizard:
Lol, I can't use bash-preexec *and* a custom debug trap unless I write a wrapper with conditional behavior and...that's why we use bash-preexec: so we get the benefit of each other's improvements.

I'm not the first one to want to be able to use debug hooks along with bash-preexec.

gaelicWizard:
Alsö, #109 would run the preexec hook multiple times on purpose.

gaelicWizard:
I'm closing this PR and will open a whole new one based on #109, which does 98% of what I need. It requires Bash version 5 or higher, but @cornfeedhobo made a comment that makes it not need v5 so I'm just doing that.

PR #120, opened 2021-10-08
Add Fig to list of tools using bash-preexec in production

mschrage:
Fig uses `bash-preexec` in our [shell integration](https://github.com/withfig/config). Thanks so much for writing this script!


rcaloras:
Thanks for the PR! No prob. Will check out out Fig!

PR #123, opened 2022-01-20
Remove the underscore prefix of `__bp_imported`

### LINKED ISSUE #122: official way to detect whether bash-preexec is loaded
1. `[[ "$(declare -p preexec_functions)" =~ "declare -a" ]] 2>/dev/null`
 2. `[[ -n ${__bp_imported} ]]`

Concern with 1 is that the test suite doesn’t guarantee existence of variable `preexec_functions`; note that in a fresh Zsh, this variable is undeclared. Concern with 2 is that I’m not sure whether this is part of the public API (given the `__bp_` prefix).

FranklinYu:
This formally exposes it as part of the public API. RFC: @dimo414.

Fixes #122.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,15 @@ export __bp_enable_subshells="true"
 ```
 This is disabled by default due to buggy situations related to to `functrace` and Bash's `DEBUG trap`. See [Issue #25](https://github.com/rcaloras/bash-preexec/issues/25)
 
+## Library authors
+If you want to detect bash-preexec in your library (for example, to add hooks to `preexec_functions` when available), use the Bash variable `${bp_imported}`:
+
+```bash
+if [[ "${bp_imported:-}" == "defined" ]]; then
+    echo "Bash-preexec is loaded."
+fi
+```
+
 ## Tests
 You can run tests using [Bats](https://github.com/bats-core/bats-core).
 ```bash
--- a/README.md
+++ b/README.md
@@ -92,10 +92,10 @@ export __bp_enable_subshells="true"
 This is disabled by default due to buggy situations related to to `functrace` and Bash's `DEBUG trap`. See [Issue #25](https://github.com/rcaloras/bash-preexec/issues/25)
 
 ## Library authors
-If you want to detect bash-preexec in your library (for example, to add hooks to `preexec_functions` when available), use the Bash variable `${bp_imported}`:
+If you want to detect bash-preexec in your library (for example, to add hooks to `preexec_functions` when available), use the Bash variable `bash_preexec_imported`:
 
 ```bash
-if [[ "${bp_imported:-}" == "defined" ]]; then
+if [[ -n "${bash_preexec_imported:-}" ]]; then
     echo "Bash-preexec is loaded."
 fi
 ```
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -91,6 +91,15 @@ export __bp_enable_subshells="true"
 ```
 This is disabled by default due to buggy situations related to to `functrace` and Bash's `DEBUG trap`. See [Issue #25](https://github.com/rcaloras/bash-preexec/issues/25)
 
+## Library authors
+If you want to detect bash-preexec in your library (for example, to add hooks to `preexec_functions` when available), use the Bash variable `${bp_imported}`:
```
rcaloras, line 95: Don't think the `$` and brackets are necessary here. I'd just have this say `bp_imported`.

rcaloras (changes requested):
Thanks for the PR @FranklinYu! Looks like there's a couple other places that use the original name:
```bash
[rcaloras:~/git/bash-preexec] master ± git grep __bp_imported
...
test/include-test.bats:  __bp_imported="defined"
test/include-test.bats:  unset __bp_imported
```
Can you also the test case for this in to use the new variable name?

bash-preexec.sh
```diff
@@ -38,10 +38,14 @@ if [[ -z "${BASH_VERSION:-}" ]]; then
...
 
 # Avoid duplicate inclusion
-if [[ "${__bp_imported:-}" == "defined" ]]; then
+if [[ "${bp_imported:-}" == "defined" ]]; then
     return 0
 fi
-__bp_imported="defined"
+bp_imported="defined"
+
+# WARNING: This variable has never been designed as part of the public API. Bash
```
dimo414, line 46: How about just making this a deprecation warning, e.g. "This variable is no longer used and should not be relied upon, use bp_imported instead."?

README.md
```diff
@@ -91,6 +91,15 @@ export __bp_enable_subshells="true"
...
+
+```bash
+if [[ "${bp_imported:-}" == "defined" ]]; then
```
dimo414, line 98: How about simply `-n`? I don't think the value matters.

bash-preexec.sh
```diff
@@ -38,10 +38,14 @@ if [[ -z "${BASH_VERSION:-}" ]]; then
 fi
 
 # Avoid duplicate inclusion
-if [[ "${__bp_imported:-}" == "defined" ]]; then
+if [[ "${bp_imported:-}" == "defined" ]]; then
     return 0
 fi
-__bp_imported="defined"
+bp_imported="defined"
```
dimo414, line 44: This is also our opportunity to use a clearer name, rather than just stripping the `__`. `bash_preexec_imported`, for instance. @rcaloras in case you have any naming preferences.

rcaloras: 👍 agree with suggested naming.

FranklinYu:
All of the suggestion above looks good to me, but apparently the test is currently broken. For example, the “include test” use `__bp_preexec_and_precmd_install` to decide whether bash-preexec is loaded, but this function isn’t there regardless. Shall I start another PR to fix the test before moving forward with this one?

FranklinYu:
I’m glad that the CI is working, and it caught a bug. I’ll fix it before marking this as “ready for review”.

FranklinYu:
I finally addressed all the comments. The history seems a bit messy; maybe consider merge by squashing the commits.

rcaloras (approved):
LGTM 👍

PR #124, opened 2022-01-21
Fix bugs in the code and in tests 

FranklinYu:
Locally verified on macOS, with Bats from MacPorts.

### DIFF
--- a/test/include-test.bats
+++ b/test/include-test.bats
@@ -3,7 +3,7 @@
 @test "should not import if it's already defined" {
   __bp_imported="defined"
   source "${BATS_TEST_DIRNAME}/../bash-preexec.sh"
-  [ -z $(type -t __bp_preexec_and_precmd_install) ]
+  [ -z $(type -t __bp_install) ]
 }
 
 @test "should import if not defined" {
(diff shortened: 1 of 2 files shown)


test/include-test.bats
```diff
@@ -3,7 +3,7 @@
 @test "should not import if it's already defined" {
   __bp_imported="defined"
   source "${BATS_TEST_DIRNAME}/../bash-preexec.sh"
-  [ -z $(type -t __bp_preexec_and_precmd_install) ]
```
FranklinYu, line 6: This bug makes test always pass, leading to false negative. For example, changing `__bp_imported` to `__bp_imported_foobar` should have failed the test, but it doesn’t.

FranklinYu:
Please let me know if you also want me to set up GitHub Action (to better verify the fixes above).

test/include-test.bats
rcaloras: Ah man, looks like this was broken in 4744d7e6fd105e. Fixed the function below then, but missed this one :/ Good fix 👍

rcaloras:
LGTM. Thanks @FranklinYu!

> Please let me know if you also want me to set up GitHub Action (to better verify the fixes above).

@FranklinYu That'd be great if you have time! I've wanted to at some point. Feel free to put together that change.

PR #125, opened 2022-01-22
Configure the first workflow for GitHub Actions

FranklinYu:
After this is verified to work, we can probably remove the Travis CI configuration.

### DIFF
--- a/.github/workflows/bats.yaml
+++ b/.github/workflows/bats.yaml
@@ -1,0 +1,17 @@
+name: Bats tests
+on: [push, pull_request]
+env:
+  functional_test: 'true'
+
+jobs:
+  test:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Install Bats
+        run: |
+          apt-get update
+          apt-get install --assume-yes bats
+      - name: Check out repository
+        uses: actions/checkout@v2
+      - name: Run tests
+        run: bats test
--- a/.github/workflows/bats.yaml
+++ b/.github/workflows/bats.yaml
@@ -1,7 +1,5 @@
 name: Bats tests
 on: [push, pull_request]
-env:
-  functional_test: 'true'
 
 jobs:
   test:


.github/workflows/bats.yaml
```diff
@@ -0,0 +1,17 @@
+name: Bats tests
+on: [push, pull_request]
+env:
+  functional_test: 'true'
```
FranklinYu, line 4: Note: I don’t know what this environment variable was for, but this variable exist in the Travis CI configuration.

.github/workflows/bats.yaml
```diff
@@ -0,0 +1,17 @@
...
+
+jobs:
+  test:
+    runs-on: ubuntu-20.04
```
FranklinYu, line 8: Alternatively `ubuntu-latest` will keep at the latest image; I personally prefer explicit versions, to have more control over when exactly to migrate. Your project, your choice.

.github/workflows/bats.yaml
```diff
@@ -0,0 +1,17 @@
...
+    steps:
+      - name: Install Bats
+        run: |
+          apt-get update
+          apt-get install --assume-yes bats
```
FranklinYu, line 13: Note: This will install [the same version](https://packages.ubuntu.com/source/focal/bats) as the previous Travis CI configuration. I think installing from APT is easier than clone-and-make. By the way, I chose `--assume-yes` over `-y` for the sake of readability (codes are written once, read many times).

.github/workflows/bats.yaml
rcaloras: Lets remove it. This was copy pasta from another configuration and thought there might be a use case that never materialized.

rcaloras (approved):
LGTM. Feel free to remove the functional_test env on this PR or in a follow up.

dimo414:
This looks great to me (so +1 for merging as-is), but it would be nice to have coverage of multiple Bash versions. Here's the [approach](https://github.com/dimo414/bash-cache/blob/master/.github/workflows/ci.yml) I tool for bash-cache (with a custom Docker image). There's also an official [bats/bats](https://hub.docker.com/r/bats/bats) image which appears to [support different versions](https://github.com/bats-core/bats-core/blob/429e86417b52f09d491778bb3cd14eaa1d25462f/.github/workflows/tests.yml#L124-L141), though it seems odd to me that they build the image for every run.

FranklinYu:
Yes, that sounds like a viable follow-up. Meanwhile I plan to add macOS Big Sur as another platform.

dimo414:
It's probably sufficient to just test against Bash 3.2, which is what macOS is stuck on.

rcaloras (approved):

rcaloras:
Merged, fixed permissions for install (https://github.com/rcaloras/bash-preexec/commit/b1630c76020416ccaaf56ecfd3f1f066dbd2461e), updated Readme (https://github.com/rcaloras/bash-preexec/commit/3e563be4f143366dc21d795d311da506be6fb75c)

PR #128, opened 2022-01-27
Try to better handle external modification to `$PROMPT_COMMAND`

gaelicWizard:
Some libraries do things like `PROMPT_COMMAND="$PROMPT_COMMAND ; _new_addition"`, that's a space-semicolon-space, and others do `PROMPT_COMMAND="_other_thing; $PROMPT_COMMAND"`. This leads to two consecutive semicolons if both libraries load after bp first sets `$PROMPT_COMMAND`. It would be counter-productive to try to account for all possibilities, so just swap the install string for a no-op. Then, try to clean up the no-op if it's all that's left.

### DIFF
--- a/bash-preexec.sh
+++ b/bash-preexec.sh
@@ -307,14 +307,16 @@ __bp_install() {
     local existing_prompt_command
     # Remove setting our trap install string and sanitize the existing prompt command string
     existing_prompt_command="${PROMPT_COMMAND:-}"
-    existing_prompt_command="${existing_prompt_command//$__bp_install_string[;$'\n']}" # Edge case of appending to PROMPT_COMMAND
-    existing_prompt_command="${existing_prompt_command//$__bp_install_string}"
+    # Edge case of appending to PROMPT_COMMAND
+    existing_prompt_command="${existing_prompt_command//$__bp_install_string/:}" # no-op
+    existing_prompt_command="${existing_prompt_command//$'\n':$'\n'/$'\n'}" # remove known-token only
+    existing_prompt_command="${existing_prompt_command//$'\n':;/$'\n'}" # remove known-token only
     __bp_sanitize_string existing_prompt_command "$existing_prompt_command"
 
     # Install our hooks in PROMPT_COMMAND to allow our trap to know when we've
     # actually entered something.
     PROMPT_COMMAND=$'__bp_precmd_invoke_cmd\n'
-    if [[ -n "$existing_prompt_command" ]]; then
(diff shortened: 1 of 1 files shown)


gaelicWizard:
I fixed the failing test

bash-preexec.sh
```diff
@@ -307,14 +307,16 @@ __bp_install() {
     local existing_prompt_command
     # Remove setting our trap install string and sanitize the existing prompt command string
     existing_prompt_command="${PROMPT_COMMAND:-}"
-    existing_prompt_command="${existing_prompt_command//$__bp_install_string[;$'\n']}" # Edge case of appending to PROMPT_COMMAND
-    existing_prompt_command="${existing_prompt_command//$__bp_install_string}"
+    # Edge case of appending to PROMPT_COMMAND
+    existing_prompt_command="${existing_prompt_command//$__bp_install_string/:}" # no-op
+    existing_prompt_command="${existing_prompt_command//$'\n':$'\n'/$'\n'}" # remove known-token only
+    existing_prompt_command="${existing_prompt_command//$'\n':;/$'\n'}" # remove known-token only
     __bp_sanitize_string existing_prompt_command "$existing_prompt_command"
```
akinomyoga, line 314: I guess these reductions of `:` can be put inside `__bp_sanitize_string`

bash-preexec.sh
```diff
@@ -307,14 +307,16 @@ __bp_install() {
...
 
     # Install our hooks in PROMPT_COMMAND to allow our trap to know when we've
     # actually entered something.
     PROMPT_COMMAND=$'__bp_precmd_invoke_cmd\n'
-    if [[ -n "$existing_prompt_command" ]]; then
+    if [[ "${existing_prompt_command:-:}" != ":" ]]; then
```
akinomyoga, line 319: Also the reduction of `existing_prompt_command=':'` into `existing_prompt_command=''` could also be put inside `__bp_sanitize_string`, so that we do not have to consider the case for `:`.

akinomyoga (commented):
@rcaloras Could you review the change in this PR (#128)? I would like to partially reuse the code in this PR, so I would like to wait for this to be merged first if this PR. More specifically, I would like to use `__bp_sanitize_string` with the reductions introduced in this PR.

**Edit**: The following suggestions may not be included in the PR. In such a case, I would include the changes in the next PR.

akinomyoga (approved):

rcaloras:
@akinomyoga thanks for reviewing. Will merge in for you!
@gaelicWizard thanks for the original PR! 👍 🍟

Issue #116, opened 2021-07-06
`__bp_preexec_invoke_exec "$_"` printed when running multiple commads


clupasq:
I originally discovered this when using [bash-it](https://github.com/Bash-it/bash-it), and posted an issue there as well: https://github.com/Bash-it/bash-it/issues/1895
I'm reproducing the relevant bits:

<kbd>Ctrl</kbd> + <kbd>X</kbd>, <kbd>Ctrl</kbd> + <kbd>E</kbd> allows opening the current bash command in the default editor.
I sometimes to this to enter multiple commands (separated by newline).
After saving and exiting the editor, the commands are run as expected.

For the examples below I'll consider the following commands:
```
echo 1
echo 2
echo 3
```

## Expected Behavior

What is expected is that only the commands and their results are being printed to the console.
Here's an example of such an interaction without bash-it:

```
$> echo # <== at this point I fired up the editor to type in the aforementioned command(s).
echo 1
1
echo 2
2
echo 3
3
```

## Current Behavior

Here's what's being printed when using bash-it:

```
❯ echo # <== using the editor to type in the commands again
echo 1
__bp_preexec_invoke_exec "$_"
1
echo 2
__bp_preexec_invoke_exec "$_"
2
echo 3
__bp_preexec_invoke_exec "$_"
3
```

I'm not sure if this is a bug; it may be the intended behavior, so please confirm.

## Steps to Reproduce

1. Press <kbd>Ctrl</kbd> + <kbd>X</kbd>, <kbd>Ctrl</kbd> + <kbd>E</kbd> on the command line
2. Use the editor to enter more than one command (newline separated)
3. Save && exit
4. Inspect the output. You're likely to see the expected output, plus some `__bp_preexec_invoke_exec "$_"` lines between each command and its output.

## Environment

* Bash-it version used: dev (git SHA: 2444a57 on 2021-06-10T16:04:56+03:00)
* List of enabled plugins, themes and aliases (use ``bash-it show (plugins/themes/aliases)``): no plugins, no aliases, powerline-multiline theme (also tested with themes bobby & nwinkler and got the same result)
* ``bash-it doctor`` output:

* Bash version: 5.0.3(1)-release 
* Operating System and version: Debian 10 (buster)

akinomyoga:
This is a bug of Bash <= 5.0, which happens when the command is executed through `fc` or <kbd>C-x C-e</kbd>. It was fixed in Bash 5.1.

[...]

Issue #104, opened 2020-02-07
preexec is called / not called in weird moments


mkielar:
I have single `preexec` and `precmd` functions defined.

```
preexec() { 
    echo "preexec for $@"
}
precmd() {
    echo "precmd for $@"
}
```
When I open new bash window, I get this:

```
precmd for
preexec for <last command from history>
14:08 ● /home/mkielar
→ ls
file1 file2
precmd for

14:09 ● /home/mkielar
→ ls
preexec for ls
file1 file2
precmd for

14:09 ● /home/mkielar
→ ls
```
So, it is called when I open new bash session, it;s not being called for first command in that bash session, and then it's being called again for any subsequent commands....

sullivan-sean:
Has anyone found a solution for this? I'm also running into this issue running `bash 5.1.8(1)-release` and `bash-preexec v0.4.1`

rcaloras:
@sullivan-sean same exactly as described here? And bash-preexec.sh is the last thing included in your bash configuration?

slash-dev:
I also have a similar problem. preexec is called when I open the terminal then it's never called again. It used to but since few months it's not working anymore.

rcaloras:
@slash-dev could you share what your debug trap and PROMPT_COMMAND look like: e.g.  running on my local machine

```bash
[rcaloras:~] 3s $ trap && echo "$PROMPT_COMMAND"
trap -- '__bp_preexec_invoke_exec "$_"' DEBUG
__bp_precmd_invoke_cmd
__lp_set_prompt
__bp_interactive_mode
```

slash-dev:
Here is the result:

```
$ trap && echo "$PROMPT_COMMAND"
__bp_precmd_invoke_cmd
__bp_interactive_mode
```

When I add debug into my precmd and preexec functions and open a new terminal I get:

```
precmd
preexec
✔ 11:50:06 (ø) 
$ 
precmd
✔ 11:50:08 (1s982ms) 
$ 
precmd
✔ 11:50:09 (2s) 
$ ls
...
precmd
✔ 11:51:02 (55s) 
$ 
```

Thank you for the quick response!

Issue #100, opened 2019-10-09
Preexec runs before accepting the command line when using readline shell command bindings.


curusarn:
Readline allows us to bind arbitrary shell commands to key sequences using `bind -x`. Using these shell commands results in `preexec()` being executed before accepting the command line instead of after. Plus `preexec()` is executed with the same argument twice effectively skipping the next accepted command-line.

### Example
- Define `preexec()` and `precmd()`
```sh
(ins)[simon@simon-pc ~]$ preexec() { echo "just typed $1"; }
(ins)[simon@simon-pc ~]$ precmd() { echo "printing the prompt"; }
printing the prompt
```
- Run some commands to see normal behavior
```sh
(ins)[simon@simon-pc ~]$ echo first
just typed echo first
first
printing the prompt
(ins)[simon@simon-pc ~]$ echo second
just typed echo second
second
printing the prompt
```
- Bind a simple shell command to `Control+X`
```sh
(ins)[simon@simon-pc ~]$ bind -x '"\C-x": READLINE_LINE="echo typed by readline"'
just typed bind -x '"\C-x": READLINE_LINE="echo typed by readline"'
printing the prompt
```
- Run normal command
```sh
(ins)[simon@simon-pc ~]$ echo third
just typed echo third
third
printing the prompt
```
- Now press `Control+X`
We can see that `preexec()` was run before we accepted the command line. We can also notice that `preexec()` run with the same argument as moments before.
```sh
just typed echo third
(ins)[simon@simon-pc ~]$ echo typed by readline
```
- Now press `Enter` to accept the command typed by readline
We can see that `preexec()` wasn't executed this time effectively skipping the command-line typed by readline.
```sh
typed by readline
printing the prompt
(ins)[simon@simon-pc ~]$ 
```

curusarn:
This issue is very similar to this one: https://github.com/rcaloras/bash-preexec/issues/27

dimo414:
Do you have any thoughts for how bash-preexec could better handle this situation? I don't think a consensus was reached in #27.

curusarn:
My immediate thought was to use some of the variables that are set when readline widgets are executed (`READLINE_LINE`, `READLINE_POINT`). The problem is that these are not set when the trap is triggered.

I have read the bash manual to look for other suitable variables but I couldn't find any.

I needed to solve this in my project. (https://github.com/curusarn/resh)
I did work around this by re-enabling bash-preexec in my own custom readline widgets/functions.
[...]

Issue #25, opened 2016-06-24
Background subshells may finalize session


davefx:
After loading bash-preexec.sh, if I launch, for example:
`
$ ( ls ) &
`
most of the times, bash exits (not always).

rcaloras:
@davefx thanks for reporting. Just recognizing this now as a result of some issues with bashhub that's using bash-preexec. Believe the changes made to support subshells in [0.3.0](https://github.com/rcaloras/bash-preexec/releases/tag/0.3.0) cause this. You can use 0.2.3 if you don't mind losing subshell support.

rcaloras:
@davefx Could you also post the version of bash you're using?
```bash
echo "$BASH_VERSION"
```

rcaloras:
Believe this is impacting bash versions >= 4.2

davefx:
I've reproduced this problem with bash 4.3.42(1) and 4.3.30(1).

As some other tips:
If I execute something after the background process, bash doesn't finish .
`
$ (ls)& echo Hello
`

rcaloras:
Was able to boil this down to a simple code block to recreate. Still not sure the best way to handle this

```bash
# Causes login shells to logout
# Look like bash versions > 4.2.46
set -o functrace > /dev/null 2>&1
no_op() { :;}
trap 'no_op' DEBUG;
# Any command in a subshell and background
( pwd ) & 
```

rcaloras:
@davefx was this occurring on a box you were ssh'd into? I can't recreate this locally regardless of the version. I can only recreate when ssh'd onto a box.

davefx:
No. This happens also locally in X terminals in my Ubuntu 16.04 system with bash 4.3.46(1)-release.

rcaloras:
Looks like this got mentioned to some folks looking into this as a bug in bash. http://lists.gnu.org/archive/html/bug-bash/2016-09/msg00006.html

adggit:
When the DEBUG trap is set with functrace on, it causes other problems too.  E.g., certain command groups piped to a pager program will get stopped (via SIGTTOU/SIGTTIN when the pager accesses the tty):

```
set -o functrace
trap ':' DEBUG

# The following gets stopped:
{ date; who | sort; } | more

[1]+  Stopped                 { date; who | sort; } | more

# But this works fine:
{ who | sort; date; } | more

```
In general (with functrace on and the DEBUG trap set) a group of command pipelines in the form below will get stopped:

```
[...]

PR #129, opened 2022-02-19
Request for public/stable APIs to enable/disable the hooks and to manually invoke registered functions

akinomyoga:
`bash-preexec` relies on the `DEBUG` trap (for `preexec`) and `PROMPT_COMMAND` (for `precmd`). In particular, the hook using the `DEBUG` trap is a kind of hack and fragile. The `DEBUG` trap is called for every single shell command (of the top-level when `set -T` or `shopt -s extdebug` is not set), which makes the execution slower. The `DEBUG` trap can interfere with other Bash configurations that use the `DEBUG` trap. The implementation of `preexec` by the `DEBUG` trap also causes problems like #115, #104, #116, #100, #25, and #6. The hook using `PROMPT_COMMAND` (for `precmd`) is not so stable, in particular when other configurations add settings such as `PROMPT_COMMAND="${PROMPT_COMMAND+$PROMPT_COMMAND;}__new_settings__"` (see #128 for example).

Therefore, when the other framework provides proper support for the `preexec` and `precmd` hooks, it would be better to just use them instead of relying on fragile `DEBUG` and `PROMPT_COMMAND`. To allow users or other Bash configuration frameworks to choose how  `precmd` and `preexec` of `bash-preexec` will be invoked, I would like to request public/stable APIs to install/uninstall `DEBUG` and `PROMPT` hooks in arbitrary timing and also the public/stable APIs to invoke `{precmd,preexec}_functions` from external mechanisms of `preexec` and `precmd`.

The public APIs that I would like to request in this PR are:

- A function to invoke `preexec_functions` (see the shell function [`bash_preexec_invoke_preexec_functions`]() in this PR)
- A function to invoke `precmd_functions` (see the shell function [`bash_preexec_invoke_precmd_functions`]() in this PR)
- A string that can be evaluated to install the hooks in `DEBUG` and `PROMPT_COMMAND` (see the shell variable [`bash_preexec_install_string`]() in this PR, which was originally named `__bp_install_string`)
- A trap string that `bash-preexec` registers to the `DEBUG` trap (see the shell variable [`bash_preexec_trapdebug_string`]())
- A function to remove the hooks in `DEBUG` and `PROMPT_COMMAND` (see the shell function [`bash_preexec_uninstall`]())

### About the API names

I have initially created commit fe46a97 with the existing naming convention of `__bp_*`. However, now we have merged #123 where `bash_preexec_*` has been adopted as the name for the public interfaces, so I have followed the new convention in commit 01bfc4e. I can adjust these naming if there are any requests.

### Background

In my project of Bash configuration ([ble.sh](https://github.com/akinomyoga/ble.sh)), I have so far received issues caused by the interference with `bash-preexec` several times. I have been trying to make the Bash configuration work with `bash-preexec`, but there is still a limitation, and it is hard to make it work perfectly (including the other issues that are reported in this upstream repository). `ble.sh` actually provides proper support for the `PRECMD` and `PREEXEC` hooks, so I would like to recommend users to use them instead of `bash-preexec`, but there are many existing settings and configurations that rely on `bash-preexec`. Also, I don't want to unnecessarily require users to create dependency on ble.sh. So I decided to rewrite the hooks by `bash-preexec` when `ble.sh` detects it. Currently, `ble.sh` assumes many things about the internals of `bash-preexec` in rewriting the hooks by `bash-preexec`, which is not ideal. What I actually need is just several public/stable APIs just I described above.

Related discussions:
- https://github.com/akinomyoga/ble.sh/issues/96#issuecomment-1019432190
- https://github.com/Bash-it/bash-it/pull/1884
- https://github.com/akinomyoga/ble.sh/issues/33
- https://github.com/akinomyoga/ble.sh/issues/176#issuecomment-1042424832


dimo414:
> `ble.sh` actually provides proper support for the `PRECMD` and `PREEXEC` hooks

Can you provide some pointers for this behavior? What is more "proper" about your approach?

akinomyoga:
> What is more "proper" about your approach?

Maybe the word choice *proper* was not the appropriate one. I'm not a native speaker so don't know what would be the best-fit word for this case, but ble.sh is an alternative implementation of a line editor. Everything including the prompt calculation and command execution is under the control of `ble.sh`. In particular, because the prompt calculation and the user commands are directly executed by ble.sh, `PRECMD` and `PREEXEC` can just be directly inserted in the execution sequence of ble.sh. I used *proper* to mean that they are designed for those purpose from the beginning unlike the `DEBUG` trap and `PROMPT_COMMAND` that are originally designed for other purposes, i.e., for `bashdb` and the prompt settings.

> Can you provide some pointers for this behavior?

It depends on what kind of information you need, but if you need the information on the interface of these hooks, it is explained in [§1.2.2 of the manual](https://github.com/akinomyoga/ble.sh/wiki/Manual-%C2%A71-Introduction#user-content-fn-blehook). If you need the implementation details, it is hard to point one place because ble.sh implements an entire line editor and requires reading larger code sections to understand the whole execution flow, but `PRECMD` is executed from [here](https://github.com/akinomyoga/ble.sh/blob/5015cb56075119d59c051a9852db035b8bd3ed12/src/edit.sh#L1572) and `PREEXEC` is executed from [here](https://github.com/akinomyoga/ble.sh/blob/5015cb56075119d59c051a9852db035b8bd3ed12/src/edit.sh#L6255). [`ble-edit/exec:gexec/invoke-hook-with-setexit`](https://github.com/akinomyoga/ble.sh/blob/5015cb56075119d59c051a9852db035b8bd3ed12/src/edit.sh#L6155-L6161) sets `$?` and `$_` to the ones by the previous user command and `$BASH_COMMAND` to the one by the previous (current) user command for `PRECMD` (`PREEXEC`). `blehook/invoke` is implemented [here](https://github.com/akinomyoga/ble.sh/blob/5015cb56075119d59c051a9852db035b8bd3ed12/src/util.sh#L1858-L1874). When a `PREEXEC` handler is a function name or a command name, the current user command that we are trying to run now is specified to the first argument.

akinomyoga:
I have added code comments to explain the new functions and force-pushed.

<details><summary>The overall diff of the forced push is here.</summary>

```diff
diff --git a/bash-preexec.sh b/bash-preexec.sh
index f1758d3..06db478 100644
--- a/bash-preexec.sh
+++ b/bash-preexec.sh
@@ -153,6 +153,9 @@ __bp_precmd_invoke_cmd() {
     bash_preexec_invoke_precmd_functions "$__bp_last_ret_value" "$__bp_last_argument_prev_command"
 }

+# This function invokes every function defined in our function array
+# "precmd_function".  This function receives the arguments $1 and $2 for $? and
+# $_, respectively, that will be set for the precmd functions.
 bash_preexec_invoke_precmd_functions() {
     local __lastexit=$1 __lastarg=$2
     # Invoke every function defined in our function array.
@@ -267,8 +270,12 @@ __bp_preexec_invoke_exec() {
     __bp_set_ret_value "$preexec_ret_value" "$__bp_last_argument_prev_command"
 }

-# This function invokes every function defined in our function array.  This
-# function assigns the last error exit status to the variable
+# This function invokes every function defined in our function array
+# "preexec_function".  This function receives the arguments $1 and $2 for $?
+# and $_, respectively, that will be set for the preexec functions.  The third
+# argument $3 specifies the user command that is going to be executed
+# (corresponding to BASH_COMMAND in the DEBUG trap).  This function assigns the
+# last non-zero exit status from the preexec functions to the variable
 # `preexec_ret_value`.  If there is no error, preexec_ret_value is set to `0`.
 bash_preexec_invoke_preexec_functions() {
     local __lastexit=$1 __lastarg=$2 __this_command=$3
@@ -367,6 +374,7 @@ __bp_install_after_session_init() {
     PROMPT_COMMAND+=${bash_preexec_install_string}
 }

+# Remove hooks installed in the DEBUG trap and PROMPT_COMMAND.
 bash_preexec_uninstall() {
     # Remove __bp_install hook from PROMPT_COMMAND
     if [[ ${PROMPT_COMMAND-} == *"$bash_preexec_install_string"* ]]; then
@@ -389,6 +397,7 @@ bash_preexec_uninstall() {
         fi
     fi
 }
+declare -ft bash_preexec_uninstall

 # Run our install so long as we're not delaying it.
 if [[ -z "${__bp_delay_install:-}" ]]; then
```
</details>

akinomyoga:
@rcaloras @dimo414 Can you also review this PR? I also think #128 and #129 should be considered. Thanks.

rcaloras:
@akinomyoga Thanks for the PR and support of this project. I'll take a look and chime in before end of week!

bash-preexec.sh
```diff
@@ -143,15 +150,22 @@ __bp_precmd_invoke_cmd() {
...
     fi
     local __bp_inside_precmd=1
+    bash_preexec_invoke_precmd_functions "$__bp_last_ret_value" "$__bp_last_argument_prev_command"
+}
 
+# This function invokes every function defined in our function array
+# "precmd_function".  This function receives the arguments $1 and $2 for $? and
+# $_, respectively, that will be set for the precmd functions.
+bash_preexec_invoke_precmd_functions() {
+    local __lastexit=$1 __lastarg=$2
```
rcaloras, line 160: Don't think the underscore naming scheme is necessary here.

bash-preexec.sh
```diff
@@ -319,10 +345,13 @@ __bp_install() {
     fi;
     PROMPT_COMMAND+='__bp_interactive_mode'
 
-    # Add two functions to our arrays for convenience
-    # of definition.
-    precmd_functions+=(precmd)
-    preexec_functions+=(preexec)
+    # Add two functions to our arrays for convenience of definition only when
+    # the functions have not yet added.
+    if [[ ! ${__bp_installed_convenience_functions-} ]]; then
```
rcaloras, line 350: why is this check necessary? Are there instances where `__bp_install` is being called twice?

bash-preexec.sh
```diff
@@ -342,8 +371,33 @@ __bp_install_after_session_init() {
     if [[ -n "$sanitized_prompt_command" ]]; then
         PROMPT_COMMAND=${sanitized_prompt_command}$'\n'
     fi;
-    PROMPT_COMMAND+=${__bp_install_string}
+    PROMPT_COMMAND+=${bash_preexec_install_string}
+}
+
+# Remove hooks installed in the DEBUG trap and PROMPT_COMMAND.
+bash_preexec_uninstall() {
```
rcaloras, line 378: I'm not convinced that bash-preexec should be responsible for maintaining any over head of uninstalling itself from both trap and prompt_command. Any libraries using these shared variables have more or less had to clean them up and maintain them on their own, as does bash-preexec with its installation. With the defined global variables what's stopping from just implementing this in your own library?

bash-preexec.sh
```diff
@@ -342,8 +371,33 @@ __bp_install_after_session_init() {
...
+    local q="'" Q="'\''"
+    if [[ $(trap -p DEBUG) == "trap -- '${bash_preexec_trapdebug_string//$q/$Q}' DEBUG" ]]; then
+        if [[ ${__bp_trap_string-} ]]; then
+            eval -- "$__bp_trap_string"
+        else
+            trap - DEBUG
+        fi
+    fi
 }
+declare -ft bash_preexec_uninstall
```
rcaloras, line 400: any reason this needs to be declared? just good practice?

rcaloras (changes requested):

rcaloras:
@akinomyoga thanks again for the PR and trying to help make bash-preexec better. This small library has gotten more widely used across other projects/companies than ever originally anticipated. As such, it's definitely opinionated toward original use cases and has clashed withe other libraries over shared state like `PROMPT_COMMAND` and the debug trap. The biggest expectation is that bash-preexec is the last modifier of these things during bash configuration and that it will be responsible for managing them. I can see how this would be an issue with your library that could effectively supersede the need for bash-preexec in general. 

I think I'm open to most of this change provided it's not breaking. Left comments and questions in line.

> By the way, what is the minimal Bash version that bash-preexec supports? 

I think 3.1 sounds right. Unless there's some other feature being used that would date it later. We should add this to the documentation at the top.

bash-preexec.sh
```diff
@@ -60,7 +60,14 @@ __bp_inside_precmd=0
 __bp_inside_preexec=0
 
 # Initial PROMPT_COMMAND string that is removed from PROMPT_COMMAND post __bp_install
-__bp_install_string=$'__bp_trap_string="$(trap -p DEBUG)"\ntrap - DEBUG\n__bp_install'
+bash_preexec_install_string=$'__bp_trap_string="$(trap -p DEBUG)"\ntrap - DEBUG\n__bp_install'
+
+# WARNING: This variable is no longer used and should not be relied upon.
+# Use ${bash_preexec_install_string} instead.
+__bp_install_string="${bash_preexec_install_string}"
```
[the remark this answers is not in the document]
akinomyoga: OK! I can remove it. I have just followed how we have preserved `__bp_imported` which was replaced by `bash_preexec_imported`. I guess the variable `__bp_install_string` is unlikely to have been used by external libraries (except for `ble.sh`) unlike `__bp_imported`, so I guess the older name can be removed.

bash-preexec.sh
```diff
@@ -244,32 +258,44 @@ __bp_preexec_invoke_exec() {
...
+
+# This function invokes every function defined in our function array
+# "preexec_function".  This function receives the arguments $1 and $2 for $?
+# and $_, respectively, that will be set for the preexec functions.  The third
+# argument $3 specifies the user command that is going to be executed
+# (corresponding to BASH_COMMAND in the DEBUG trap).  This function assigns the
+# last non-zero exit status from the preexec functions to the variable
+# `preexec_ret_value`.  If there is no error, preexec_ret_value is set to `0`.
+bash_preexec_invoke_preexec_functions() {
+    local __lastexit=$1 __lastarg=$2 __this_command=$3
```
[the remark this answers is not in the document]
akinomyoga: OK, I will follow the existing local variables used.

I have prefixed these new variables with underscores in case the hook functions want to modify the global variables of the conflicting names, but if we would care about it, the same would also apply to the existing variable names. I decided to just follow the existing ones.

bash-preexec.sh
```diff
@@ -244,32 +258,44 @@ __bp_preexec_invoke_exec() {
         return
     fi
 
-    # Invoke every function defined in our function array.
+    local preexec_ret_value
```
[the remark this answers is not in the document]
akinomyoga: The variable is set by `bash_preexec_invoke_preexec_functions`.

Now I'm afraid that you might not be familiar with the Bash's variable scoping. Bash's variable scope is dynamic scoping, so the callee sees all the local variables in the caller chain. You can see the following examples.

```bash
$ f2() { var=1; }
$ f1() { local var=0; f2; echo "$var"; }
$ f1
```

Using this dynamic scoping, we can pass/receive data through local variables between callers and callees without affecting the global context in Bash.

OK. Now I understood why you have asked about the underscore naming. Because of the dynamic scoping, the local variables defined in the function affect the context of all the callees, so we need to care about the name conflicts even if it is a local variable.

bash-preexec.sh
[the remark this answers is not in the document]
akinomyoga: As I have commented in https://github.com/rcaloras/bash-preexec/pull/129#discussion_r859137813, these local variables can affect the context of callers (i.e., the hook functions stored in the array `preexec_functions` by the user).

bash-preexec.sh
akinomyoga: Yes, when the user or another library would like to temporarily turn off bash-preexec and then later want to re-enable it. `ble.sh` would do that.

bash-preexec.sh
akinomyoga: We need it because the effect of `trap - DEBUG` is local to that function call unless the trace attribute is set to the function.

bash-preexec.sh
rcaloras: Curious, what's the reason ble.sh would re-enable it?

bash-preexec.sh
akinomyoga: > With the defined global variables what's stopping from just implementing this in your own library?

If the setup of `bash-preexec.sh` would never change in the future forever, I could just assume the current setup of `bash-preexec.sh` and write a code to clean up the things that are installed by the current `bash-preexec.sh`.

However, if the `__bp_install` configuration can change in the future (e.g. as suggested in #128), the cleanup code also needs to be updated. I don't think it is a good idea to update the cleanup code at the users' side every time the `__bp_install` configuration of `bash-preexec.sh` is changed. In particular, if the method of uninstalling the hooks would not be provided by `bash-preexec.sh`, the user/library needs to maintain the separate cleanup code for each of *all* the past versions of `bash-preexec.sh` and to detect the precise version of `bash-preexec.sh`.

This is actually the reason that I would like to request the "stable" public API rather than relying on the details of the internal implementation that can be frequently changed.

bash-preexec.sh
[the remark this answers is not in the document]
akinomyoga: c7b9867

bash-preexec.sh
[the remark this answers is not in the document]
akinomyoga: dda9193

bash-preexec.sh
akinomyoga: dda9193

bash-preexec.sh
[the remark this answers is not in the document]
akinomyoga: a2216fc I have changed the strategy as suggested in the thread https://github.com/rcaloras/bash-preexec/pull/129#discussion_r858662337.

bash-preexec.sh
```diff
@@ -244,32 +258,44 @@ __bp_preexec_invoke_exec() {
...
+# argument $3 specifies the user command that is going to be executed
+# (corresponding to BASH_COMMAND in the DEBUG trap).  This function assigns the
+# last non-zero exit status from the preexec functions to the variable
+# `preexec_ret_value`.  If there is no error, preexec_ret_value is set to `0`.
+bash_preexec_invoke_preexec_functions() {
+    local __lastexit=$1 __lastarg=$2 __this_command=$3
     local preexec_function
     local preexec_function_ret_value
-    local preexec_ret_value=0
+    preexec_ret_value=0
```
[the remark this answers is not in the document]
akinomyoga: a2216fc

bash-preexec.sh
akinomyoga: 01527e5 I have added a code comment.

bash-preexec.sh
akinomyoga: ble.sh allows a user to temporarily turn off ble.sh by `ble-detach`. (For example, I have received some reports from a user that there is a performance issue with ble.sh in a directory with tens of thousands of files so the user temporarily turns off ble.sh while the current working directory is in that directory.) In the detached state, ble.sh would like to re-enable the original `bash-preexec.sh` hook because ble.sh does not provide the feature in the detached state.

akinomyoga:
Thank you for your careful review and valuable comments! I have addressed or commented on the points that you have raised in each review comment.

> > By the way, what is the minimal Bash version that bash-preexec supports?
> 
> I think 3.1 sounds right. Unless there's some other feature being used that would date it later. We should add this to the documentation at the top.

Thank you for the clarification! Maybe we might also add the version check at the beginning of `bash-preexec.sh` (just after the existing check for the variable `BASH_VERSION`).

rcaloras:
> Thank you for the clarification! Maybe we might also add the version check at the beginning of bash-preexec.sh (just after the existing check for the variable BASH_VERSION).

Sure. Feel free to add in separate commit if you'd like. Not necessary in this stack.

bash-preexec.sh
```diff
@@ -144,21 +147,38 @@ __bp_precmd_invoke_cmd() {
...
+# "precmd_function".  This function receives the arguments $1 and $2 for $? and
+# $_, respectively, that will be set for the precmd functions. This function
+# returns the last non-zero exit status of the hook functions. If there is no
+# error, this function returns 0.
+bash_preexec_invoke_precmd_functions() {
+    local lastexit=$1 lastarg=$2
     # Invoke every function defined in our function array.
     local precmd_function
+    local precmd_function_ret_value
+    local precmd_ret_value=0
```
rcaloras, line 165: This is new functionality right? can we add a test for it on this new function. Not sure if it's covered by the existing coverage.

bash-preexec.sh
```diff
@@ -256,7 +296,7 @@ __bp_preexec_invoke_exec() {
         # Only execute each function if it actually exists.
         # Test existence of function with: declare -[fF]
         if type -t "$preexec_function" 1>/dev/null; then
-            __bp_set_ret_value ${__bp_last_ret_value:-}
+            __bp_set_ret_value "$lastexit" "$lastarg"
```
rcaloras, line 299: Is there a reason to add the lastarg value here? This wasn't being set prior no?

bash-preexec.sh
rcaloras: Can we add a small test to cover this?

bash-preexec.sh
rcaloras: I get your reasoning and desire for this as a way to help with your library and not just implementing it yourself. I just think there's no real standard here I've seen other libraries that have clashed with bash-preexec as well since they also make the assumption they'll have access to these parts of the system configuration. ([Liquid Prompt](https://github.com/nojhan/liquidprompt) is one example) My assumption is, these too can cause issues with ble.sh and we're trying to chase individual fires here.  

I am concerned that this is an esoteric use case and will likely just add operational overhead to maintaining it. With that said, if any other contributors see value in this I'm supportive.

rcaloras (changes requested):

rcaloras:
@dimo414 or any other contributors have thoughts on the necessity of an uninstall function i.e. the proposed`bash_preexec_uninstall`?

bash-preexec.sh
akinomyoga: Yes! Thank you for the suggestion! I have added the test in commit e846228.

bash-preexec.sh
akinomyoga: Thanks. Maybe I miss your exact point (What does "the system configuration" mean?), but I think I'll later need to look at Liquid Prompt.

bash-preexec.sh
akinomyoga: Yes! I have added new tests for these new functions 918b451.

bash-preexec.sh
akinomyoga: Oh, OK. Right. The last argument `$_` has not been set for `preexec` before.

I have added it for consistency with `precmd_functions`. It seems to be introduced for `precmd` in d056f7dc37e30724f09eb7487a060311caf49990 via PR #32. Is there a reason that we do not provide `preexec` functions with `$_` while  we provide `precmd` functions with `$_`? As a side note, similar functionalities in ble.sh, blehook PRECMD/PREEXEC/POSTEXEC, all set `$_` for the hook functions.

In PR #32, Zsh's behavior is referenced for `precmd`. I have now checked Zsh's behavior for `preexec`. Zsh seems to set `$_` also for `preexec`:

```zsh
$ zsh
% preexec1() { echo "preexec1: lastarg='$_'"; }
% preexec2() { echo "preexec2: lastarg='$_'"; }
% add-zsh-hook preexec preexec1
% add-zsh-hook preexec preexec2
preexec1: lastarg='preexec1'
% echo hello
preexec1: lastarg='preexec2'
preexec2: lastarg='preexec2'
hello
% echo world
preexec1: lastarg='hello'
preexec2: lastarg='hello'
world
```

akinomyoga:
I have noticed that an existing test for `__bp_install` is actually broken:

https://github.com/rcaloras/bash-preexec/blob/d30880087e5eed62f629f6fd7d7c61cd27f3dceb/test/bash-preexec.bats#L53-L59

Here, line 58 should be `trap - DEBUG` instead of `trap DEBUG` because the install string has been updated in 
e36de177c081fdfdf3f448716364cfbcafc26a2e for PR #106. It is an oversight in e36de17.

However, now I believe we can and should use `bash_preexec_install_string` / `__bp_install_string`. I have updated the test to use `bash_preexec_install_string` (4a37150).

rcaloras:
@akinomyoga thanks for updating with tests. Looks good. Will hold for any other feedback/input from other contributors.

akinomyoga:
There was a conflict, so I rebased and squashed commits.

Issue #132, opened 2022-06-23
ShellCheck highlights several issues with the file - Shellcheck could be disabled, or issues resolved


jamie-wearsafe:
It can be disabled by adding this line to the shell file:

```bash
# shellcheck disable=SC2292,SC2250,SC2086,SC1087,SC2248,SC2001,SC1007,SC2034
```
Issues can be seen by adding the script to the [shellcheck online editor](https://www.shellcheck.net/):
<img width="1279" alt="image" src="https://user-images.githubusercontent.com/52926764/175314677-408eb38d-3901-42dd-b292-3dc6212ba3c9.png">

dimo414:
Feel free to send a PR addressing the ShellCheck findings, but it would be good to limit the disable directives to specific offending lines where possible.

rcaloras:
Should be addressed in #135

PR #135, opened 2022-09-11
Update to remove shellcheck warnings

adamprice2:
Hi,
I went ahead and fixed the shellcheck warnings mentioned in https://github.com/rcaloras/bash-preexec/issues/132

Changes to actual code were minimal (couple of quotes and a newline I think).

Test suite runs correctly on my system.
 bats test
 ✓ sourcing bash-preexec should exit with 1 if we're not using bash 
 ✓ __bp_install should exit if it's already installed 
 ✓ __bp_install should remove trap logic and itself from PROMPT_COMMAND 
 ✓ __bp_install should preserve an existing DEBUG trap 
 ✓ __bp_sanitize_string should remove semicolons and trim space 
 ✓ Appending to PROMPT_COMMAND should work after bp_install 
 ✓ Appending or prepending to PROMPT_COMMAND should work after bp_install_after_session_init 
 ✓ Adding to PROMPT_COMMAND before and after initiating install 
 ✓ Adding to PROMPT_COMMAND after with semicolon 
 ✓ during install PROMPT_COMMAND and precmd functions should be executed each once 
 ✓ No functions defined for preexec should simply return 
 ✓ precmd should execute a function once 
 ✓ precmd should set $? to be the previous exit code 
 ✓ precmd should set $BP_PIPESTATUS to the previous $PIPESTATUS 
 ✓ precmd should set $_ to be the previous last arg 
 ✓ preexec should execute a function with the last command in our history 
 ✓ preexec should execute multiple functions in the order added to their arrays 
 ✓ preecmd should execute multiple functions in the order added to their arrays 
 ✓ preexec should execute a function with IFS defined to local scope 
 ✓ precmd should execute a function with IFS defined to local scope 
 ✓ preexec should set $? to be the exit code of preexec_functions 
 ✓ in_prompt_command should detect if a command is part of PROMPT_COMMAND 
 ✓ __bp_adjust_histcontrol should remove ignorespace and ignoreboth 
 ✓ preexec should respect HISTTIMEFORMAT 
 ✓ preexec should not strip whitespace from commands 
 ✓ preexec should preserve multi-line strings in commands 
 ✓ preexec should work on options to 'echo' commands 
 ✓ should not import if it's already defined 
 ✓ should import if not defined 
 ✓ bp should stop installation if HISTTIMEFORMAT is readonly 

30 tests, 0 failures

### DIFF
--- a/bash-preexec.sh
+++ b/bash-preexec.sh
@@ -32,6 +32,9 @@
 #  using: the "DEBUG" trap, and the "PROMPT_COMMAND" variable. If you override
 #  either of these after bash-preexec has been installed it will most likely break.
 
+# Tell shellcheck what kind of file this is.
+# shellcheck shell=bash
+
 # Make sure this is bash that's running and return otherwise.
 # Use POSIX syntax for this line:
 if [ -z "${BASH_VERSION-}" ]; then
--- a/bash-preexec.sh
+++ b/bash-preexec.sh
@@ -245,8 +245,7 @@ __bp_preexec_invoke_exec() {
     local this_command
     this_command=$(
         export LC_ALL=C
-        # shellcheck disable=SC1007
-        HISTTIMEFORMAT= builtin history 1 | sed '1 s/^ *[0-9][0-9]*[* ] //'
+        HISTTIMEFORMAT='' builtin history 1 | sed '1 s/^ *[0-9][0-9]*[* ] //'
     )
 
     # Sanity check to make sure we have something to invoke our function with.
--- a/bash-preexec.sh
+++ b/bash-preexec.sh
@@ -291,7 +291,8 @@ __bp_install() {
     trap '__bp_preexec_invoke_exec "$_"' DEBUG
 
     # Preserve any prior DEBUG trap as a preexec function
-    local prior_trap=$(sed "s/[^']*'\(.*\)'[^']*/\1/" <<<"${__bp_trap_string:-}")
+    local prior_trap
+	prior_trap=$(sed "s/[^']*'\(.*\)'[^']*/\1/" <<<"${__bp_trap_string:-}")
     unset __bp_trap_string
     if [[ -n "$prior_trap" ]]; then
         eval '__bp_original_debug_trap() {
--- a/bash-preexec.sh
+++ b/bash-preexec.sh
@@ -138,8 +138,8 @@ __bp_interactive_mode() {
 __bp_precmd_invoke_cmd() {
     # Save the returned value from our last command, and from each process in
     # its pipeline. Note: this MUST be the first thing done in this function.
-	# BP_PIPESTATUS may be unused, ignore
-	# shellcheck disable=SC2034
+    # BP_PIPESTATUS may be unused, ignore
+    # shellcheck disable=SC2034
 
     __bp_last_ret_value="$?" BP_PIPESTATUS=("${PIPESTATUS[@]}")
 
@@ -245,7 +245,7 @@ __bp_preexec_invoke_exec() {
     local this_command
     this_command=$(
         export LC_ALL=C
-		# shellcheck disable=SC1007
+        # shellcheck disable=SC1007
         HISTTIMEFORMAT= builtin history 1 | sed '1 s/^ *[0-9][0-9]*[* ] //'
     )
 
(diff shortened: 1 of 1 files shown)


bash-preexec.sh
```diff
@@ -239,6 +245,7 @@ __bp_preexec_invoke_exec() {
     local this_command
     this_command=$(
         export LC_ALL=C
+		# shellcheck disable=SC1007
         HISTTIMEFORMAT= builtin history 1 | sed '1 s/^ *[0-9][0-9]*[* ] //'
```
akinomyoga, line 249: Does this work?

```suggestion
        HISTTIMEFORMAT='' builtin history 1 | sed '1 s/^ *[0-9][0-9]*[* ] //'
```

bash-preexec.sh
```diff
@@ -284,7 +291,10 @@ __bp_install() {
     trap '__bp_preexec_invoke_exec "$_"' DEBUG
 
     # Preserve any prior DEBUG trap as a preexec function
-    local prior_trap=$(sed "s/[^']*'\(.*\)'[^']*/\1/" <<<"${__bp_trap_string:-}")
+    local prior_trap
+    # we can't easily do this with variable expansion. Leaving as sed command.
+	# shellcheck disable=SC2001
+	prior_trap=$(sed "s/[^']*'\(.*\)'[^']*/\1/" <<<"${__bp_trap_string:-}")
```
akinomyoga, line 297: If we want to eliminate the use of `sed`, we may use the regular expression through `[[ ... ]]`.

```suggestion
    local reg="'(.*)'"
    [[ ${__bp_trap_string:-} =~ $reg ]]
    local prior_trap=${BASH_REMATCH[1]-${__bp_trap_string:-}}
```

bash-preexec.sh
```diff
@@ -310,6 +320,7 @@ __bp_install() {
     local existing_prompt_command
     # Remove setting our trap install string and sanitize the existing prompt command string
     existing_prompt_command="${PROMPT_COMMAND:-}"
+	# shellcheck disable=SC1087
     existing_prompt_command="${existing_prompt_command//$__bp_install_string[;$'\n']}" # Edge case of appending to PROMPT_COMMAND
```
akinomyoga, line 324: I think what `shellcheck` recommends is this:

```suggestion
    existing_prompt_command="${existing_prompt_command//${__bp_install_string}[;$'\n']}" # Edge case of appending to PROMPT_COMMAND
```

akinomyoga (commented):
I'm not a maintainer, so please take these as random comments. `bash-preexec` uses four spaces for one level of indentation, but the submitted PR uses inconsistent indentation by a tab character. I think you should update the indentation.

bash-preexec.sh
adamprice2: It passes the tests and is syntactically the same, I see no reason why it shouldn't work.
In fact the reason for the change is that this is clearer.

Explicitly setting HISTTIMEFORMAT to the empty string '' seems clearer than the old text to me.

bash-preexec.sh
adamprice2: Makes sense, agree.

bash-preexec.sh
adamprice2: If this works then feel free to submit a PR, not something I have time to test though.

bash-preexec.sh
akinomyoga: > you can't nest curly braces like that,

What do you mean? We can do that.

> the maintainer can update the code if they feel it is needed.

Right. The maintainer can finally decide whether we ignore the warning or follow the suggestion.

bash-preexec.sh
akinomyoga: OK. The original `shellcheck` warning implied a possibility of reducing the spawning cost of the external command `sed`. The suggestion by `shellcheck` doesn't actually work in this case, so I just showed another way of reducing the cost of `sed`.

bash-preexec.sh
adamprice2: In fact you are right, 
TIL that you can use bracketed variable names inside a parameter expansion. 
OK Thanks for the new knowledge.

rcaloras: We can put through this PR as is for now. 

I'm in favor of the update though vs disabling the warning. Feel free to submit this update as a small PR as well if it is essentially a no-op.

rcaloras:
@adamprice2 thanks for the PR and contribution!
@akinomyoga Thanks for your continued input and great feedback on this project. 

Feel free to submit further PRs for any of the other small updates here as mentioned.

rcaloras:
LGTM 👍

PR #136, opened 2022-09-11
Check Bash version 3.1+

akinomyoga:
As discussed in https://github.com/rcaloras/bash-preexec/pull/129#issuecomment-1046169131 and https://github.com/rcaloras/bash-preexec/pull/129#issuecomment-1109971595, the current codebase relies on Bash 3.1 features. In this PR, I mention it in README and check the Bash version in the beginning of the script.

### DIFF
--- a/bash-preexec.sh
+++ b/bash-preexec.sh
@@ -41,6 +41,12 @@ if [ -z "${BASH_VERSION-}" ]; then
     return 1;
 fi
 
+# We only support Bash 3.1+.
+# Note: BASH_VERSINFO is first available in Bash-2.0.
+if [[ -z "${BASH_VERSINFO-}" || BASH_VERSINFO[0] -lt 3 || BASH_VERSINFO[0] -eq 3 && BASH_VERSINFO[1] -lt 1 ]]; then
+    return 1
+fi
+
 # Avoid duplicate inclusion
 if [[ -n "${bash_preexec_imported:-}" ]]; then
     return 0
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 Bash-Preexec 
 ============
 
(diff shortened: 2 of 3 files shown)


bash-preexec.sh
```diff
@@ -38,6 +38,12 @@ if [ -z "${BASH_VERSION-}" ]; then
     return 1;
 fi
 
+# We only support Bash 3.1+.
+# Note: BASH_VERSINFO is first available in Bash # 2.0.
+if [[ -z "${BASH_VERSINFO[0]}" || BASH_VERSINFO[0] -lt 3 || BASH_VERSINFO[0] -eq 3 && BASH_VERSINFO[1] -lt 1 ]]; then
+    return 1
```
rcaloras, line 44: Can this effectively replace the check above on L37? If so lets just remove the other check seems to supersede it.

akinomyoga: This version check relies on `[[` and the Bash arrays, so I'd like to test whether `[[` and arrays can be used prior to the version check. Ah, but the Bash arrays are only available from bash-2.0. I'll adjust it.

akinomyoga: I have slightly updated it so that the array existence is checked (`bash >= 3.0`) before using the array `BASH_VERSINFO`.

akinomyoga: Actually, there is a way to combine L37 and L43: test the Bash version in POSIX syntax as in the following code:

```bash
if [ -z "${BASH_VERSINFO-}" ] || [[ 'BASH_VERSINFO[0]' -lt 3 ]] || { [[ 'BASH_VERSINFO[0]' -eq 3 ]] && [[ 'BASH_VERSINFO[1]' -lt 1 ]]; }; then
  return
fi
```

I think we want to minimize the use of the POSIX syntax and also this is not so readable, so I prefer to separate the test for Bash (in POSIX syntax) and the test for the Bash version (in Bash syntax). Nevertheless, if you would like to combine them I can update the PR. What do you think?

rcaloras: Sure. Makes sense.

akinomyoga:
Thanks!

Issue #121, opened 2021-11-01
Tests fail with bats 1.5.0


rycee:
See below for a test run with Bats 1.5.0 (the same run succeeds with Bats 1.4.1). Originally found in a [NixOS CI job](https://hydra.nixos.org/build/156972814/nixlog/2).

``` sh-session
$ bats test
 ✓ __bp_install_after_session_init should exit with 1 if we're not using bash
 ✓ __bp_install should exit if it's already installed
 ✓ __bp_install should remove trap logic and itself from PROMPT_COMMAND
 ✓ __bp_install should preserve an existing DEBUG trap
 ✓ __bp_sanitize_string should remove semicolons and trim space
 ✓ Appending to PROMPT_COMMAND should work after bp_install
 ✓ Appending or prepending to PROMPT_COMMAND should work after bp_install_after_session_init
 ✓ Adding to PROMPT_COMMAND before and after initiating install
 ✓ Adding to PROMPT_COMMAND after with semicolon
 ✓ during install PROMPT_COMMAND and precmd functions should be executed each once
 ✓ No functions defined for preexec should simply return
 ✓ precmd should execute a function once
 ✓ precmd should set $? to be the previous exit code
 ✓ precmd should set $BP_PIPESTATUS to the previous $PIPESTATUS
 ✓ precmd should set $_ to be the previous last arg
 ✓ preexec should execute a function with the last command in our history
 ✓ preexec should execute multiple functions in the order added to their arrays
 ✓ preecmd should execute multiple functions in the order added to their arrays
 ✗ preexec should execute a function with IFS defined to local scope
   (in test file test/bash-preexec.bats, line 260)
     `[ $status -eq 0 ]' failed
   Error: Test file "/home/rycee/devel/nixpkgs/source/merge" does not exist
 ✗ precmd should execute a function with IFS defined to local scope
   (in test file test/bash-preexec.bats, line 270)
     `[ $status -eq 0 ]' failed
   Error: Test file "/home/rycee/devel/nixpkgs/source/merge" does not exist
 ✓ preexec should set $? to be the exit code of preexec_functions
 ✓ in_prompt_command should detect if a command is part of PROMPT_COMMAND
 ✓ __bp_adjust_histcontrol should remove ignorespace and ignoreboth
 ✓ preexec should respect HISTTIMEFORMAT
 ✓ preexec should not strip whitespace from commands
 ✓ preexec should preserve multi-line strings in commands
[...]

PR #138, opened 2022-11-17
Fix a test for #131

akinomyoga:
This fixes an issue reported in https://github.com/rcaloras/bash-preexec/issues/121#issuecomment-1318281281 by @rycee:

> Thanks, I tester just now with bats 1.8.2 and curiously I get
> 
> ```
> ok 19 preexec should execute a function with IFS defined to local scope
> not ok 20 precmd should execute a function with IFS defined to local scope
> # (in test file test/bash-preexec.bats, line 272)
> #   `[ $status -eq 0 ]' failed
> ok 21 preexec should set $? to be the exit code of preexec_function
> ```

This is an oversight of updating the test for change #131.


akinomyoga:
Sorry, this is already fixed in the master. I withdraw this PR.

Issue #140, opened 2023-03-06
Functions called in opposite order in midnight commander subshell


josef-kyrian:
I have defined simple functions for test and added to .bashrc. They print command to run and it's return code
```bash
preexec() { echo "RUN: $1"; }
precmd() { echo "EXIT: $?"; }
source ~/.bash-preexec.sh
```

In bash shell it's working ok:
```
# ls /
RUN: ls /
bin  boot  cdrom  dev  etc  home  initrd.img  lib  lib32  lib64  lxc  media  mnt  opt  proc  root  run  sbin  snap  srv  sys  tmp  usr  var  vmlinuz  www
EXIT: 0
```

But in midnight commander subshell functions are called in opposite order
```
# ls /
bin  boot  cdrom  dev  etc  home  initrd.img  lib  lib32  lib64  lxc  media  mnt  opt  proc  root  run  sbin  snap  srv  sys  tmp  usr  var  vmlinuz  www
EXIT: 0
RUN: ls /
```

What's wrong? Thanks for help

akinomyoga:
`bash-preexec` doesn't support the case where `PROMPT_COMMAND` is modified after the shell startup and the first prompt. However, Midnight Commander installs its hook by *directly inputting `PROMPT_COMMAND+=...`, etc. as user commands* from stdin after starting the shell.

More specifically, `bash-preexec` requires that all the other `PROMPT_COMMAND` hooks are enclosed between `__bp_precmd_invoke_cmd` and `__bp_interactive_mode`. This is normally achieved through the initial `DEBUG` trap (`__bp_install`) installed by `bash-preexec`. However, Midnight Commander injects codes after the shell startup and modifies `PROMPT_COMMAND`. This results in the following value of `PROMPT_COMMAND`:

```bash
$ declare -p PROMPT_COMMAND
declare -- PROMPT_COMMAND=$'__bp_precmd_invoke_cmd\n__bp_interactive_mode\npwd>&11;kill -STOP $$'
```

The substring `pwd>&11;kill -STOP $$` is the hook installed by `mc`. We can see that the `mc` hook string is not enclosed within `__bp_precmd_invoke_cmd` and `__bp_interactive_mode`. Instead, this should look like the following:

```bash
PROMPT_COMMAND=$'__bp_precmd_invoke_cmd\npwd>&11;kill -STOP $$\n__bp_interactive_mode'
```

I'm not sure if there is a solution at the user side.

PR #141, opened 2023-03-06
Support the array PROMPT_COMMAND (bash >= 5.1)

akinomyoga:
This is a fix for the issue I reported in https://github.com/rcaloras/bash-preexec/issues/130#issuecomment-1455852628.

ShellCheck seems to have started to complain about `-lt`, `-eq`, etc. in the conditional command `[[ ... ]]`, so I also included a commit (77aec5ff212daed4f0471264ee844c00cdaecc8d) to workaround the shellcheck warning.

### DIFF
--- a/bash-preexec.sh
+++ b/bash-preexec.sh
@@ -183,8 +183,8 @@ __bp_set_ret_value() {
 
 __bp_in_prompt_command() {
 
-    local prompt_command_array
-    IFS=$'\n;' read -rd '' -a prompt_command_array <<< "${PROMPT_COMMAND:-}"
+    local prompt_command_array IFS=$'\n;'
+    read -rd '' -a prompt_command_array <<< "${PROMPT_COMMAND[*]:-}"
 
     local trimmed_arg
     __bp_trim_whitespace trimmed_arg "${1:-}"
@@ -290,7 +290,7 @@ __bp_preexec_invoke_exec() {
 
 __bp_install() {
     # Exit if we already have this installed.
-    if [[ "${PROMPT_COMMAND:-}" == *"__bp_precmd_invoke_cmd"* ]]; then
+    if [[ "${PROMPT_COMMAND[*]:-}" == *"__bp_precmd_invoke_cmd"* ]]; then
         return 1;
     fi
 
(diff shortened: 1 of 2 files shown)


bash-preexec.sh
```diff
@@ -332,11 +332,14 @@ __bp_install() {
...
-        PROMPT_COMMAND+=${existing_prompt_command}$'\n'
-    fi;
-    PROMPT_COMMAND+='__bp_interactive_mode'
+    PROMPT_COMMAND='__bp_precmd_invoke_cmd'
+    PROMPT_COMMAND+=${existing_prompt_command:+$'\n'$existing_prompt_command}
+    if ((BASH_VERSINFO[0] > 5 || BASH_VERSINFO[0] == 5 && BASH_VERSINFO[1] >= 1)); then
+        PROMPT_COMMAND+=('__bp_interactive_mode')
+    else
+        # shellcheck disable=SC2179 # PROMPT_COMMAND is not an array in bash <= 5.0
+        PROMPT_COMMAND+=$'\n__bp_interactive_mode'
```
akinomyoga, line 341: ShellCheck seems to believe `PROMPT_COMMAND` is an array even in `bash < 5.1` because of the above line `PROMPT_COMMAND+=(...)` for `bash >= 5.1`, so ShellCheck suggests rewriting it to `PROMPT_COMMAND+=($'\n__bp_interactive_mode')`. I disabled this warning SC2179 (and related warnings below, SC2178 and SC2179) because this is a wrong suggestion.

akinomyoga:
I decided to separate a PR for a possible solution to #140 because it would conflict with #128.

bash-preexec.sh
```diff
@@ -43,7 +43,7 @@ fi
 
 # We only support Bash 3.1+.
 # Note: BASH_VERSINFO is first available in Bash-2.0.
-if [[ -z "${BASH_VERSINFO-}" || BASH_VERSINFO[0] -lt 3 || BASH_VERSINFO[0] -eq 3 && BASH_VERSINFO[1] -lt 1 ]]; then
```
dimo414, line 46: I'd need to dig through the revision history to be sure, but I suspect this intentionally doesn't use arithmetic expansions. Either way, it would be better to make this change separately from this PR.

akinomyoga: Thanks! I have introduced the version check in #136. At that time, I just used the single conditional command `[[ ... ]]` because it looked simpler to me compare to combining `[[ ... ]] || (( ... ))`. This time, I have checked the updated version check works as expected with Bash 1.14.7, 2.0, 3.0, and 3.1. Bash 1.14.7 doesn't have BASH_VERSINFO, so it is rejected by `[[ -z "${BASH_VERSINFO-}" ]]`. Bash 2.0+ has an arithmetic command `(( ... ))` and seems to work as expected.

Anyway, I'd separate the PR!

akinomyoga: I dropped the commit and created a separate PR at #142. Thank you.

akinomyoga:
I have rebased the PR on top of the latest master (with #128 included).

rcaloras:
👍 thanks for the PR @akinomyoga! Merging.

akinomyoga:
Thanks!

Issue #153, opened 2024-02-18
preexec doesn't work for no-op command `:`


JeffFaer:
This might fall into this kind of a situation, and I apologize in advance: https://xkcd.com/1172/

----

I have a habit of running `:` specifically to execute preexecs. Today, I updated my version of bash-preexec, and that functionality no longer works.

1. https://github.com/rcaloras/bash-preexec/blob/1f77dc09356b58b5e2877c1e87158c48c4c377bc/bash-preexec.sh#L330

    When bash-preexec installs itself it replaces its install command with a no-op command `:`. It does not always manage to remove that no-op from the prompt command

2. https://github.com/rcaloras/bash-preexec/blob/1f77dc09356b58b5e2877c1e87158c48c4c377bc/bash-preexec.sh#L245
   
    bash-preexec does not execute preexec commands if it looks like it's executing something from the `$PROMPT_COMMAND`

I think it should be as simple as adding another prune attempt when we're processing `existing_prompt_command`. I'll put something together

PR #154, opened 2024-02-18
Handle another edge case in $PROMPT_COMMAND.

JeffFaer:
In particular: bash-preexec installs itself by appending __bp_install_string to $PROMPT_COMMAND. That means that there's a pretty good change that __bp_install_string will be the last thing in $PROMPT_COMMAND when __bp_install is invoked.

So there's a pretty good change that the $existing_prompt_command logic in __bp_install will result in the pre-existing prompt command followed by `:` and no trailing newline. That situation is not cleaned up by the two existing special cases.

Fix https://github.com/rcaloras/bash-preexec/issues/153


akinomyoga:
This is a part of #143. In particular, https://github.com/rcaloras/bash-preexec/pull/143/commits/fe5aedb3368d78fc47e017ceffca1d8265cdef0d is the commit for colon reduction with other refactoring and also contains the unit tests for that.

PR #160, opened 2024-07-11
Do not enable `bash-preexec.sh` in non-interactive shells

akinomyoga:
From the discussion in #159:

> > Is it ok to say that `bash-preexec.sh` is designed to be able to run outside of the interactive mode as well?
> 
> As far as I can tell, `bash-preexec.sh` doesn't target the non-interactive mode. The concept of the hooks `preexec` and `precmd` is specific to the interactive use of the shell. Also, even if `bash-preexec.sh` is sourced, I think nothing would happen because `PROMPT_COMMAND` (which will invoke the `precmd` hook) is not active in the non-interactive shell, and the `preexec` will also never be executed because it is only enabled after a call of the `precmd` hook. In addition, the `DEBUG` trap will be called for every command, which would become a large overhead, although the trap doesn't do anything actually. Therefore, I think `bash-preexec.sh` should not be loaded in the non-interactive shell.
> 
> Maybe we can have an explicit check for the interactive shell at the beginning of `bash-preexec.sh`.


bash-preexec.sh
```diff
@@ -47,6 +47,11 @@ if [[ -z "${BASH_VERSINFO-}" ]] || (( BASH_VERSINFO[0] < 3 || (BASH_VERSINFO[0]
     return 1
 fi
 
+# We do not enable bash-preexc in non-interactive shells (except in bats tests)
+if [[ $- != *i* && ${BATS_TEST_FILENAME-} != *.bats ]]; then
```
LecrisUT, line 51: Why not make `bats` run with `-i` instead?

akinomyoga:
> Maybe it's worth making environment variable overrides. Here's a snippet from `Lmod`, and although not related to interactive shell, would be good reference:
> 
> ```shell
> if [ -z "${LMOD_ALLOW_ROOT_USE+x}" ]; then
>   LMOD_ALLOW_ROOT_USE=yes
> fi
> 
> ( [ -n "${USER_IS_ROOT:-}" ] || ( [ "${LMOD_ALLOW_ROOT_USE:-}" != yes ] && [ $(id -u) = 0 ] ) ) && return
> ```

What's that? These lines seem to be set up by `/etc/profiles.d/modules.sh` (i.e., the entry point of Lmod) and do not seem to be something each module should perform. Does Lmod instruct every module to include those lines? Also, I'm not sure if we should include such a setting specific to the specific framework in the upstream codebase. Shouldn't they be added by the Fedora packages?

bash-preexec.sh
akinomyoga: What do you mean? Bats doesn't have such an option. I also checked the upstream `bats-core/bats-core` for the latest version, but it doesn't seem to support the `-i` option either. If you mean `bash -i "$(type -P bats)" test`, it doesn't work because `bats` anyway internally spawns `bash` without `-i`.

LecrisUT:
`/etc/profiles.d/modules.sh` is loaded only once in order to define the `module` command. That section is a fallthrough to check if the file is sourced by the `root` and if there are overrides. This file is defined in `/usr/share/lmod/lmod/init/profile` and afaict, `Lmod` controls and installs that file.

> Shouldn't they be added by the Fedora packages?

Of course, that would also be possible, but maybe some standardization of override environment variables should be considered if that would be added.

bash-preexec.sh
LecrisUT: > If you mean `bash -i "$(type -P bats)" test`, it doesn't work because `bats` anyway internally spawns `bash` without `-i`.

Yeah I was thinking about that. Was also looking if there is an equivalent `set` but that also seems to not be there.

LecrisUT: I've skimmed through my `/etc/profile.d` and the only one different I could find is `colorls.sh`
```shell
[ ! -t 0 ] && return
```

akinomyoga: > I've skimmed through my `/etc/profile.d` and the only one different I could find is `colorls.sh`
> 
> ```shell
> [ ! -t 0 ] && return
> ```

I have difficulty in understanding the intent of this comment. Do you request me to do something on this PR before this is merged?

akinomyoga:
> `/etc/profiles.d/modules.sh` is loaded only once in order to define the `module` command. That section is a fallthrough to check if the file is sourced by the `root` and if there are overrides. This file is defined in `/usr/share/lmod/lmod/init/profile` and afaict, `Lmod` controls and installs that file.

I understand that part, but the question is whether `Lmod` has requirements on the behavior of *other modules*. This part seems like the loading check for the Lmod module *itself*. You've suggested it as a *good reference*, but I'm wondering about the intent for the *good reference*.

- Did you mean just an example to test the root user?
- Or did you mean a kind of template that the other modules need to follow in order to work under `Lmod`?

> > Shouldn't they be added by the Fedora packages?
>
> Of course, that would also be possible, but maybe some standardization of override environment variables should be considered if that would be added.

As far as I see your explanation, it just seems like a loader of the Lmod framework on `/etc/profiles.d/modules.sh`. Then, I think this should rather be maintained by the packages because `/etc/profile.d/*.sh` are controlled by packages but not managed in the upstream projects. This project `bash-preexec.sh` doesn't have control of `/etc/profile.d/modules.sh`.

bash-preexec.sh
LecrisUT: Just suggesting some alternatives. `bash -i` seems to not be usable within `bats`, but maybe `[ ! -t 0 ]` can achieve similar check for interactive. The only weirdness is the check against `BATS_TEST_FILENAME`, but if it can't be worked around, than it should be fine in the current approach

akinomyoga: > maybe `[ ! -t 0 ]` can achieve similar check for interactive

Maybe? How much are you sure about this? Did you actually try it?

LecrisUT: I've only tried it as:
```console
$ [ ! -t 0 ] && echo "Non-interactive" || echo "Interactive"
Interactive
```

My bash knowledge doesn't go deep, and I don't know how to setup a non-interactive environment to test that. As far as I can tell, the nuance is that `-t` checks if it is outputting to a terminal or not, and `i` flag is checking for the existence of the flag or not. Probably in Github CI, `-t` will show as "interactive" while the `i` flag will show as "non-interative", but which is the appropriate result in this case :shrug:

LecrisUT: Re Github CI, both of them seem to output as `non-interactive`, so that's irrelevant:
https://github.com/LecrisUT/test-interactive/actions/runs/11071448788/job/30763327247

akinomyoga: Thank you for the explanation. So if I understand it correctly, you just suggested `[ ! -t 0 ]` to check the interactive shell? I haven't yet tried, but I don't see how that solves the issue of `BATS_TEST_FILENAME`.

LecrisUT:
> @LecrisUT Do you request this PR to include the root user check?

Deferred, I trust your judgement on whether it is relevant to have an override for the interactive is needed or not. The root check can be addressed separately.

bash-preexec.sh
LecrisUT: Well we can try if the check gives different results with `[ ! -t 0 ]`. Depends on how `bats` is spawning the environment I guess. The explicit check for `BATS_TEST_FILENAME` feels weird, and probably having an override of `BASH_PREXEC_NO_CHECKS` which would be populated by the `bats` test would be safer. Your call on this.

Also as a non-author and external commenter I cannot resolve discussions so feel free to resolve it when you have tested the other options and decided on one.

akinomyoga: I now tried `[ ! -t 0 ]` instead of `[[ $- != *i* && ${BATS_TEST_FILENAME-} != *.bats ]]`. As I anticipated, it doesn't work when the `bats` test is triggered in an environment where `stdin` is not connected to TTY/PTY.

akinomyoga: > The explicit check for `BATS_TEST_FILENAME` feels weird,

Yeah, I actually agree with this.

> and probably having an override of `BASH_PREXEC_NO_CHECKS` which would be populated by the `bats` test would be safer.

That seems cleaner, but I'm not sure if it is possible. I'll try.

akinomyoga:
> Deferred,

Thanks.

bash-preexec.sh
akinomyoga: I confirmed that the variable set at the top level in the `.bats` files is visible from each test.

I updated the commit: [`3e1581c..30e09cb`](https://github.com/rcaloras/bash-preexec/compare/3e1581c41b1b5b0191fb75d5b8684ffdb183ae47..30e09cb48024c6ef027588757c33c6b23b3b64f8).

I first defined `BASH_PREEXEC_SOMETHING` at the top level, but I changed it to `__bp_something` because it is still not a part of a public interface and is not intended to be used by users.

LecrisUT: Awesome, no notes. Go ahead and resolve this thread.

akinomyoga: Thanks!

akinomyoga:
@jparise I'd appreciate it if you could review this PR. In https://github.com/rcaloras/bash-preexec/pull/152#issuecomment-1963036028, the maintainers seem to require several external reviewers to leave reviews before merging (except for the trivial ones like #165), though there is no sufficient audience watching this repository. Recently, I'm the only reviewer regularly watching this repository.

jparise (commented):
This code looks good to me. I think it might be nice to mention `__bp_inside_test` in `test/README.md` because users who want to test `bash-preexec.sh` as part of their own tests would also need to set it.

Overall, though, I'm not clear if this is solving a known problem or if enforcing this restriction is more of a "correctness / precondition"?

akinomyoga:
> I think it might be nice to mention `__bp_inside_test` in `test/README.md` because users who want to test `bash-preexec.sh` as part of their own tests would also need to set it.

@jparise Thanks for the suggestion! That makes sense. I updated it ([](https://github.com/rcaloras/bash-preexec/compare/6b7a9bef566f0b533926fe54a7e88d19b46d10bf..acc0aa5c9b2c9b360bbb2a8c12da1cad24cd853f)).

> Overall, though, I'm not clear if this is solving a known problem or if enforcing this restriction is more of a "correctness / precondition"?

I think even without this condition, there wouldn't be a real problem.

This is just for the cleanliness of the shell environment not to introduce random functions and variables that wouldn't be used. Some users put `source bash-preexec.sh` in their `.bashrc` or `.bash_profile` without the checking for the interactive shell, or even distributions' `/etc/profile` or `/etc/bashrc` sometimes source `bash-preexec.sh` in both interactive and non-interactive shells through `/etc/profile.d/bash-preexec.sh`.

jparise (approved):

akinomyoga:
@rcaloras Could you take a look at this too?

akinomyoga:
@rcaloras This is another PR that I think is relatively trivial and easy to review.

bash-preexec.sh
```diff
@@ -47,6 +47,12 @@ if [[ -z "${BASH_VERSINFO-}" ]] || (( BASH_VERSINFO[0] < 3 || (BASH_VERSINFO[0]
     return 1
 fi
 
+# We do not enable bash-preexc in non-interactive shells (except in tests).
+# `__bp_inside_test` is set in test/*.bats in bash-preexec repository.
+if [[ $- != *i* && ! ${__bp_inside_test-} ]]; then
```
jparise, line 52: Do you think we could use one of the [Bats special variables](https://bats-core.readthedocs.io/en/stable/writing-tests.html#special-variables) instead (e.g. `$BATS_VERSION`)?

jparise: That's a good point. Maybe use `$BATS_TEST_TAGS` and tag first-party tests with a `bash-preexec` tag? That would also allow others to opt-in.

It wouldn't be that different from what you're doing now, but it maybe a more standard (Bats-specific) approach.

akinomyoga: That seems useful. I'll later check it.

PR #161, opened 2024-07-25
Avoid checking PATH for functions

agriffis:
Searching PATH for functions is slow and unnecessary...

### DIFF
--- a/bash-preexec.sh
+++ b/bash-preexec.sh
@@ -163,8 +163,7 @@ __bp_precmd_invoke_cmd() {
     for precmd_function in "${precmd_functions[@]}"; do
 
         # Only execute this function if it actually exists.
-        # Test existence of functions with: declare -[Ff]
-        if type -t "$precmd_function" 1>/dev/null; then
+        if __bp_func_exists "$precmd_function"; then
             __bp_set_ret_value "$__bp_last_ret_value" "$__bp_last_argument_prev_command"
             # Quote our function invocation to prevent issues with IFS
             "$precmd_function"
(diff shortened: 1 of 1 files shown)


agriffis:
I force-pushed this a couple times just to clean it up.

agriffis:
Hey there, thanks for looking at this.

I did not benchmark. The motivator for this PR is a badly-behaved system where `type -t` without modifying PATH takes nearly half a second. My apologies for not realizing this would affect the common case.

FWIW, command substitutions typically don't fork in bash unless they call an external command. Bash has some internal optimizations to avoid forking for builtins, for example. You can see this with:

```
echo $$
1234

echo $(echo $$)
1234

echo $(bash -c 'echo $$')
2345
```

That's a great point about searching functions before PATH, though! I agree I opened this without sufficient research and experimentation!

akinomyoga:
> FWIW, command substitutions typically don't fork in bash unless they call an external command. Bash has some internal optimizations to avoid forking for builtins, for example. You can see this with:
> 
> ```
> echo $$
> 1234
> 
> echo $(echo $$)
> 1234
> 
> echo $(bash -c 'echo $$')
> 2345
> ```

No, you misunderstand `$$`. `$$` always returns the process ID of the main shell. To check the process ID of the current process, you need to reference `$BASHPID`. `ksh` performs such an optimization, but Bash doesn't do that except for `$(< file)` in recent versions. Please read Bash Reference Manual:

- [Special Parameters - Bash Reference Manual](https://www.gnu.org/software/bash/manual/bash.html#Special-Parameters) for `$$`
- [Bash Variables - Bash Reference Manual](https://www.gnu.org/software/bash/manual/bash.html#Bash-Variables) for `$BASHPID`

You should check the results of the following commands:

```bash
$ echo "$$:$BASHPID"
1997853:1997853
$ echo $(echo "$$:$BASHPID")
1997853:1998890
```

agriffis:
Oh, thanks!

PR #165, opened 2025-01-09
Fix indentation and remove trailing semicolons

akinomyoga:
A few lines have the 2-space indentation. We fix them in this PR.

Also, several lines have suffix semicolons. Those semicolons might have been added by e.g. `bash --pretty-print bash-preexec.sh`, but we currently don't have such a semicolon for most of the lines anyway. There are two options for consistency:

- We suffix semicolons to every line that can have a semicolon.
- Or we remove all unnecessary semicolons.

In this PR, we choose the latter, the majority.

### DIFF
--- a/bash-preexec.sh
+++ b/bash-preexec.sh
@@ -38,7 +38,7 @@
 # Make sure this is bash that's running and return otherwise.
 # Use POSIX syntax for this line:
 if [ -z "${BASH_VERSION-}" ]; then
-    return 1;
+    return 1
 fi
 
 # We only support Bash 3.1+.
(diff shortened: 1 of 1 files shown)


akinomyoga:
@rcaloras Could you also check other PRs, if possible? I think this one is a trivial fix, so it shouldn't take time to review.

Also, #160 would be another trivial one.

I also want to bring #96, which I reviewed and approved before, to your attention.

My other PRs are a bit non-trivial, so I understand that you would require some other reviewers.

I think issues #100 and #130 can be closed as resolved. #104 may also be closed because this seems to have been an external issue (or compatibility with an external setting), and the details needed to prepare any precautions, if we would do, are missing.

akinomyoga:
Thanks!

PR #169, opened 2025-01-13
Use `command` to execute `sed`

jparise:
We want our execution environment to be a predictable as possible. Similar to how we use `builtin` to ensure we're running the shell's builtin function (and not a redefinition), use `command` to ensure we're running `sed` and not a shell function or alias.

### DIFF
--- a/bash-preexec.sh
+++ b/bash-preexec.sh
@@ -300,7 +300,7 @@ __bp_install() {
     local prior_trap
     # we can't easily do this with variable expansion. Leaving as sed command.
     # shellcheck disable=SC2001
-    prior_trap=$(sed "s/[^']*'\(.*\)'[^']*/\1/" <<<"${__bp_trap_string:-}")
+    prior_trap=$(builtin command sed "s/[^']*'\(.*\)'[^']*/\1/" <<<"${__bp_trap_string:-}")
     unset __bp_trap_string
     if [[ -n "$prior_trap" ]]; then
         eval '__bp_original_debug_trap() {


jparise:
Some other notes:

- We could also use the `\sed` syntax if that's preferred.
- It would be nice to remove this `sed` call entirely, but it looks like it might be difficult to re-implement this in "native" bash.

The motivation for this change is to avoid conflicts with (incompatible) `sed` aliases: https://github.com/ghostty-org/ghostty/pull/4991

akinomyoga (approved):

bash-preexec.sh
```diff
@@ -300,7 +300,7 @@ __bp_install() {
     local prior_trap
     # we can't easily do this with variable expansion. Leaving as sed command.
     # shellcheck disable=SC2001
-    prior_trap=$(sed "s/[^']*'\(.*\)'[^']*/\1/" <<<"${__bp_trap_string:-}")
+    prior_trap=$(builtin command sed "s/[^']*'\(.*\)'[^']*/\1/" <<<"${__bp_trap_string:-}")
```
akinomyoga, line 303: FWIW, we have been simply using `command` in `bash-completion` and `oh-my-bash`, though I know kitty shell integration prefixes `builtin` to everything where applicable.

akinomyoga (changes requested):
> * It would be nice to remove this `sed` call entirely, but it looks like it might be difficult to re-implement this in "native" bash.

Actually. you could simply do

```bash
builtin eval -- "local split=(${__bp_trap_string:-})"
prior_trap=${split[2]}
```

and it's more accurate. The current version in the `master` branch is broken when the trap string contains `'`. See this:

```console
$ (trap "echo 'hello'" DEBUG; trap -p DEBUG)
hello
trap -- 'echo '\''hello'\''' DEBUG
```

bash-preexec.sh
jparise: Yes, I made that as a last minute change given that I referenced `builtin` in my commit message. I think it's unlikely someone would attempt to redefine `command`, but using `builtin command` here does make this the most explicit.

akinomyoga:
You can modify this PR. I don't care about the credit about it.

bash-preexec.sh
akinomyoga: Yeah, either works for me, but then the next question would be whether we should prefix `builtin` to everything like `builtin local`, `builtin return`, `builtin unset`, `builtin set`, `builtin shopt`, etc.

jparise:
> You can modify this PR. I don't care about the credit about it.

I created #170 so the maintainers can compare the two approaches.

akinomyoga:
> I created #170 so the maintainers can compare the two approaches.

Thanks!

akinomyoga (commented):
I think #170 is better.

akinomyoga:
@rcaloras Thanks for checking and merging PRs! Now that #170 is merged, this PR (which implements an alternative approach to #170) can be closed.

PR #170, opened 2025-01-13
Improve prior_trap processing

jparise:
We use `trap` to bootstrap our installation function (__bp_install). We remove our code upon first execution but need to restore any preexisting trap calls. We previously used `sed` to process the trap string, but that had two downsides:

1. `sed` is an external command dependency. It needs to exist on the system, and we need to invoke it in a subshell (which has some runtime cost).
2. The regular expression pattern was imperfect and didn't handle trickier cases like `'` characters in the trap string:

        $ (trap "echo 'hello'" DEBUG; trap -p DEBUG)
        hello
        trap -- 'echo '\''hello'\''' DEBUG

This change removes the dependency on `sed` by locally evaluating the trap string and extracting any prior trap. This works reliably because we control the format our trap string, which looks like this (with newlines expanded):

    __bp_trap_string="$(trap -p DEBUG)"
    trap - DEBUG
    __bp_install

### DIFF
--- a/test/bash-preexec.bats
+++ b/test/bash-preexec.bats
@@ -103,6 +103,26 @@ set_exit_code_and_run_precmd() {
   (( trap_count_snapshot < trap_invoked_count ))
 }
 
+@test "__bp_install should preserve an existing DEBUG trap containing quotes" {
+  trap_invoked_count=0
+  foo() { (( trap_invoked_count += 1 )); }
+
+  # note setting this causes BATS to mis-report the failure line when this test fails
+  trap "foo && echo 'hello' >/dev/null" debug
+  [ "$(trap -p DEBUG | cut -d' ' -f3-7)" == "'foo && echo '\''hello'\'' >/dev/null'" ]
+
+  bp_install
+  trap_count_snapshot=$trap_invoked_count
+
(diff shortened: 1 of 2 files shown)


test/bash-preexec.bats
```diff
@@ -103,6 +103,26 @@ set_exit_code_and_run_precmd() {
   (( trap_count_snapshot < trap_invoked_count ))
 }
 
+@test "__bp_install should preserve an existing DEBUG trap containing quotes" {
+  trap_invoked_count=0
+  foo() { (( trap_invoked_count += 1 )); }
+
+  # note setting this causes BATS to mis-report the failure line when this test fails
+  trap "foo && echo 'hello' >/dev/null" debug
```
jparise, line 111: The previous `sed`-based pattern would fail with more complex cases like this.

akinomyoga (approved):

jparise:
As a proof point, [Ghostty 1.1.0](https://ghostty.org/docs/install/release-notes/1-1-0) shipped with this change to its bundled bash-preexec, and we haven't heard of any issues through the pre- or post-release period yet.

akinomyoga (approved):

akinomyoga (approved):

PR #172, opened 2025-08-03
Mention Ghostty

jparise:
Also, remove Fig, which was shut down in September 2024.


akinomyoga (commented):
I didn't know Fig had been shut down!

Since you are a Ghostty contributor, it feels a bit unfair to add only Ghostty in updating the list of projects. Today, many projects are relying on bash-preexec for their Bash integration. As far as I know, Warp also includes bash-preexec. Starship, VS Code Terminal, and Atuin also rely on bash-preexec as an option. Of course, it is not a complete list. Could you make an effort to search for projects that use `bash-preexec` and add them to the list? **edit**: Also, Bash-it and Oh-My-Bash use bash-preexec.

PR #187, opened 2026-05-26
Preserve $? and $_ on __bp_precmd_invoke_cmd early returns

gnachman:
__bp_precmd_invoke_cmd saves the previous command's exit status into __bp_last_ret_value at the top and restores it via __bp_set_ret_value on the normal exit path. Two early-return paths bypass that restore:

1. The `(( __bp_inside_precmd > 0 )); then return` reentrancy guard leaves $? equal to the exit status of the `(( ... ))` test, which is always 0 when the condition is true.

2. The `return 0` inside the nested-call-frame branch (the "another framework saved our PROMPT_COMMAND" case) explicitly clobbers $? to 0.

Both paths run before reaching the trailing __bp_set_ret_value, so the next command in PROMPT_COMMAND (and any direct caller of __bp_precmd_invoke_cmd) sees $? = 0 instead of the user's actual last exit status. Add an explicit __bp_set_ret_value call (and replace `return 0` with a bare `return`) on both early-return paths so $? and $_ flow through cleanly.

Includes bats coverage for both paths. The path-2 test clears BATS_VERSION inside the nested call because the path's own guard short-circuits when BATS_VERSION is set (so bats's own scaffolding doesn't trip it).

### DIFF
--- a/bash-preexec.sh
+++ b/bash-preexec.sh
@@ -232,7 +232,7 @@ __bp_precmd_invoke_cmd() {
     # prompt command" by another precmd execution loop. This avoids infinite
     # recursion.
     if (( __bp_inside_precmd > 0 )); then
-        return
+        return "$__bp_last_ret_value"
     fi
 
     # Check and adjust PROMPT_COMMAND to make sure that PROMPT_COMMAND has the
@@ -245,7 +245,7 @@ __bp_precmd_invoke_cmd() {
             # (e.g., starship). In this case, we do not need to invoke precmd
             # because it is supposed to be already processed by the top-level
             # __bp_precmd_invoke_cmd.
-            return 0
+            return "$__bp_last_ret_value"
         fi
     fi
 
--- a/test/bash-preexec.bats
+++ b/test/bash-preexec.bats
@@ -10,8 +10,9 @@ setup() {
 
 # Evaluates all the elements of PROMPT_COMMAND
 eval_PROMPT_COMMAND() {
-  local prompt_command
+  local lastexit=$? lastarg=$_ prompt_command
   for prompt_command in "${PROMPT_COMMAND[@]}"; do
+    __bp_set_ret_value "$lastexit" "$lastarg"
     eval "$prompt_command"
   done
 }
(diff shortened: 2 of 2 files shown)


bash-preexec.sh
```diff
@@ -232,6 +232,12 @@ __bp_precmd_invoke_cmd() {
     # prompt command" by another precmd execution loop. This avoids infinite
     # recursion.
     if (( __bp_inside_precmd > 0 )); then
+        # Preserve $? and $_ across the early return: without this, the
+        # bare `return` propagates the exit status of the `(( ... ))` test
+        # (always 0 when the condition is true), clobbering the value the
+        # user code is about to inspect in PROMPT_COMMAND-time prompt
+        # substitutions.
+        __bp_set_ret_value "$__bp_last_ret_value" "$__bp_last_argument_prev_command"
         return
```
akinomyoga:
```suggestion
        return "$__bp_last_ret_value"
```

Was this PR generated by AI? If so, I wouldn't feel like writing explanations to convince the machine, but I can explain how it's wrong if you request. Also, the mention of `$_` in the added code comments should be removed.

bash-preexec.sh
```diff
@@ -245,7 +251,13 @@ __bp_precmd_invoke_cmd() {
...
             # because it is supposed to be already processed by the top-level
             # __bp_precmd_invoke_cmd.
-            return 0
+            #
+            # Preserve $? and $_ instead of forcing them to 0 — same reasoning
+            # as the early-return branch above; the top-level call will run
+            # the user's precmd_functions, and they expect to see the real
+            # last-command exit status.
+            __bp_set_ret_value "$__bp_last_ret_value" "$__bp_last_argument_prev_command"
+            return
```
akinomyoga:
```suggestion
            return "$__bp_last_ret_value"
```

akinomyoga (changes requested):

bash-preexec.sh
akinomyoga: > but I can explain how it's wrong if you request.

Actually, I think it's a good idea to make the AI explain it based on a few hints. You can provide the following hints to your AI:

Hints to AI:
- `$_` is overwritten by `return`
- `$_` is also overwritten on the function exit, so setting `$_` inside a function is meaningless
- We already call `__bp_precmd_invoke_cmd "$_"` to preserve `"$_"`.

gnachman: THanks for pointing that out. It looks like return is indeed good enough, and quite a bit simpler.

bash-preexec.sh
gnachman: Done

akinomyoga (approved):
Thanks!

akinomyoga:
For future reference, this is probably a port of the following modification.

https://gitlab.com/gnachman/iterm2/-/commit/cdaaddbe04542ebd5031e1d99627f4adad8c2b30
https://github.com/gnachman/iTerm2-shell-integration/commit/aacd1214ca95ae9e7bd397d17e67670201170053
https://gitlab.com/gnachman/iterm2/-/work_items/11016

I was interested in the actual setup that caused the early return, but this PR was probably not motivated by a case of the early return. The original version of bash-preexec embedded in the iTerm2 integration at that time didn't set `$?` even at the end of `__bp_precmd_invoke_cmd`, so `$?` was always 0. Then, all the branches were modified in the embedded version, while the upstream bash-preexec fixed only the end of the function.

